PR #85, opened 2019-07-24
super basic org details page

UnsungHero97:
Screen Shot 2019-07-26 at 3 30 20 PM

### DIFF
--- a/src/containers/orgs/OrgContainer.js
+++ b/src/containers/orgs/OrgContainer.js
@@ -5,12 +5,16 @@
 import React, { Component } from 'react';
 
 import styled from 'styled-components';
+import { faPlus } from '@fortawesome/pro-regular-svg-icons';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { Map } from 'immutable';
 import { OrganizationsApiActions } from 'lattice-sagas';
 import {
+  Button,
   Card,
   CardSegment,
   Colors,
+  Input,
   Spinner,
 } from 'lattice-ui-kit';
 import { connect } from 'react-redux';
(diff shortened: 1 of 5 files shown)


src/containers/orgs/OrgContainer.js
```diff
@@ -5,28 +5,148 @@
...
 type Props = {
   actions :{
     getOrganization :RequestSequence;
+    resetRequestState :(actionType :string) => void;
   };
-  getOrgRequestState :RequestState;
   org :Map;
   orgId :UUID;
+  requestStates :{
+    [typeof GET_ORGANIZATION] :RequestState;
```
giuliacampana: why are you storing your `requestState` as `typeof GET_ORGANIZATION`?

UnsungHero97: I'm trying a new thing here. I got tired of thinking of a variable name for the request state, and I wanted to use the action as the identifier, so this happened.

```
requestStates :{
  [typeof GET_ORGANIZATION] :RequestState;
};
```

^ this is for flow. What it means is that `requestStates` is an object with `typeof GET_ORGANIZATION` being a key in that object. If you look at how `GET_ORGANIZATION` is defined, you'll see:

```
const GET_ORGANIZATION :'GET_ORGANIZATION' = 'GET_ORGANIZATION';
```

The type of `GET_ORGANIZATION` is literally the string `GET_ORGANIZATION`, and that translates to flow resolving `typeof GET_ORGANIZATION` to the string literal `GET_ORGANIZATION`. In other words, with this, I don't have to think of and define a variable name like `getOrgRequestState`, and instead, I can reuse the existing `GET_ORGANIZATION` variable already defined and just do this:

```
// in mapStateToProps
return {
  requestStates: {
    [GET_ORGANIZATION]: state.getIn(['orgs', GET_ORGANIZATION, 'requestState']),
  },
};

// in the component
const { requestStates } = this.props;
... requestStates[GET_ORGANIZATION] ...
```

In more visual terms, this is what flow sees:

Screen Shot 2019-07-30 at 3 32 12 PM

giuliacampana: oh that's super cool! i hadn't thought much about how the `type of` the action is just the string literal of the action's name. seems like a great method to use going forward for naming request states!

PR #101, opened 2019-08-06
add / remove members

UnsungHero97:
Screen Shot 2019-08-05 at 6 57 12 PM


src/containers/orgs/OrgsSagas.js
```diff
@@ -1,36 +1,133 @@
...
+
+  try {
+    yield put(searchMembersToAddToOrg.request(action.id, action.value));
+
+    const { organizationId, query } = action.value;
+    if (!isValidUUID(organizationId)) {
+      throw new Error('organizationId must be a valid UUID');
+    }
+
+    yield delay(1000);
```
solomontang: Did you forget to remove this?

src/containers/orgs/OrgContainer.js
```diff
@@ -476,18 +559,68 @@ class OrgContainer extends Component<Props, State> {
...
+          isNonEmptyString(memberSearchQuery) && !searchResultCardSegments.isEmpty()
+            ? (
+              <div>
+                <Card>{searchResultCardSegments}</Card>
+              </div>
+            )
+            : null
+        }
+        {
+          !memberCardSegments.isEmpty()
```
solomontang: Not critical but you can short circuit instead of use a ternary
```
{
  !memberCardSegments.isEmpty()
  && (
    <div>
      <Card>{memberCardSegments}</Card>
    </div>
  )
}
```

src/containers/orgs/OrgsSagas.js
UnsungHero97: No, this is intentional. It's there alongside `takeLatest` to avoid firing off a search request for each character typed during member search. It's like a throttle.

src/containers/orgs/OrgContainer.js
UnsungHero97: Thanks!

src/containers/orgs/OrgsSagas.js
solomontang: I believe redux-sagas has first party support for this behavior
https://redux-saga.js.org/docs/api/#throttlems-pattern-saga-args

UnsungHero97: I tried throttle; it didn't work quite the way I wanted. I can demo if you'd like.

src/containers/orgs/OrgsReducer.js
```diff
@@ -87,6 +102,27 @@ export default function orgsReducer(state :Map<*, *> = INITIAL_STATE, action :Ob
...
+        REQUEST: () => state
+          .setIn([ADD_MEMBER_TO_ORGANIZATION, 'requestState'], RequestStates.PENDING)
+          .setIn([ADD_MEMBER_TO_ORGANIZATION, seqAction.id], seqAction),
+        SUCCESS: () => {
+          const storedSeqAction :SequenceAction = state.getIn([ADD_MEMBER_TO_ORGANIZATION, seqAction.id]);
+          if (storedSeqAction) {
+            const { memberId } = storedSeqAction.value;
+            return state
+              .deleteIn(['memberSearchResults', memberId])
+              .setIn([ADD_MEMBER_TO_ORGANIZATION, 'requestState'], RequestStates.SUCCESS);
```
giuliacampana: you seem to update org members after removing one but not after adding one. is there any reason for that?

UnsungHero97: Yes.

To add a member, I first have to search via the `PrincipalsApi`. What I get back as member search result is a different type than what is stored in redux. What I do instead is reload the org details on successful add:

```
if (requestStates[ADD_MEMBER_TO_ORGANIZATION] === RequestStates.SUCCESS
    && prevProps.requestStates[ADD_MEMBER_TO_ORGANIZATION] === RequestStates.PENDING) {
  actions.getOrganizationDetails(orgId);
}
```

giuliacampana: oooh i see. that does make sense!

PR #120, opened 2019-08-20
adding isOwner logic to the org details page

UnsungHero97:
This is what a user will see if they are not an owner of an org:

Screen Shot 2019-08-20 at 5 12 24 PM


UnsungHero97:
> Why does the user still see the "Delete Organization" button if they're not the owner?

Good catch!

PR #139, opened 2019-09-07
org title, description editing, org sections, refactoring

UnsungHero97:
#### Owner, Member, Public sections

Screen Shot 2019-09-06 at 5 52 00 PM

For the sake of time, I'm using fabricate to edit the org's title and description. In a future iteration, this will likely change to an inline edit, similar to `InlineEditableControl`. Also, the edit buttons will likely change to be circular. #WIP

#### Editing org title using fabricate

Screen Shot 2019-09-06 at 5 43 03 PM
Screen Shot 2019-09-06 at 5 43 34 PM
Screen Shot 2019-09-06 at 5 43 40 PM
Screen Shot 2019-09-06 at 5 43 53 PM

#### Editing org description using fabricate

Screen Shot 2019-09-06 at 4 32 39 PM
Screen Shot 2019-09-06 at 4 32 53 PM
Screen Shot 2019-09-06 at 4 33 05 PM
Screen Shot 2019-09-06 at 4 33 20 PM


src/containers/orgs/components/OrgTitleSection.js
```diff
@@ -0,0 +1,193 @@
...
+  font-size: 28px;
+  font-weight: normal;
+  margin: 0;
+  padding: 0;
+`;
+
+const FormMinusMargin = styled(Form)`
+  flex: 1;
+  margin: -30px;
+`;
```
giuliacampana: Are these styles that should be factored into `lattice-fabricate`?

UnsungHero97: `lattice-fabricate` has `StyledForm` which defines some important (and hard to understand) base formatting:

https://github.com/openlattice/lattice-fabricate/blob/develop/src/form/src/components/styled/StyledForm.js

I think the `margin: -30px` should definitely NOT happen, and `StyledForm` should be updated to reflect that. As for the `flex: 1`... meh... this one is harder to reason about.

giuliacampana: That makes sense about the margin. Ya these styles are definitely hard to undestand—Is the margin being set by RJSF? And ya, I think I agree about `flex: 1`.

UnsungHero97: It's not a margin. It's actually a padding, set here:

https://github.com/openlattice/lattice-fabricate/blob/develop/src/form/src/components/styled/StyledForm.js#L27

The selector is complicated:

```
> div.form-group.field.field-object > :first-child {
  ...
  & > * {
    padding: 30px;
  }
  ...
}
```

It can also be applied to nested children. I only want the negative margin to apply at the highest level, and I haven't had time to untangle the CSS yet. The reason is polish.

PR #200, opened 2019-10-19
basic roles page

UnsungHero97:
basic roles page
Screen Shot 2019-10-31 at 2 15 41 PM

clicking on a member shows what roles they have
Screen Shot 2019-10-31 at 2 16 04 PM

clicking on a role shows which members have that role
Screen Shot 2019-10-31 at 2 16 12 PM


solomontang:
It seems like we probably want to add some switches/toggles to LUK instead of relying on checkboxes. No change required from you for this PR. Just something for us to add. moving forward

PR #276, opened 2019-11-22
role grants + admin

UnsungHero97:


src/containers/orgs/OrgsReducer.js
```diff
@@ -155,6 +168,31 @@ export default function orgsReducer(state :Map = INITIAL_STATE, action :Object)
...
+          .setIn([ADD_CONNECTION, 'requestState'], RequestStates.PENDING)
+          .setIn([ADD_CONNECTION, seqAction.id], seqAction),
+        SUCCESS: () => {
+          const storedSeqAction :SequenceAction = state.getIn([ADD_CONNECTION, seqAction.id]);
+          if (storedSeqAction) {
+            const { connection, organizationId } = storedSeqAction.value;
+            return state
+              .setIn([ADD_CONNECTION, 'requestState'], RequestStates.SUCCESS)
+              .updateIn(
+                ['orgs', organizationId, 'connections'],
```
bergieboy: Should `requestState`, `orgs`, and `connections` be consts?

src/containers/orgs/OrgsReducer.js
```diff
@@ -710,6 +773,29 @@ export default function orgsReducer(state :Map = INITIAL_STATE, action :Object)
...
+          .setIn([UPDATE_ROLE_GRANT, 'requestState'], RequestStates.PENDING)
+          .setIn([UPDATE_ROLE_GRANT, seqAction.id], seqAction),
+        SUCCESS: () => {
+          const storedSeqAction :SequenceAction = state.getIn([UPDATE_ROLE_GRANT, seqAction.id]);
+          if (storedSeqAction) {
+            const grant :Grant = storedSeqAction.value.grant;
+            const organizationId :UUID = storedSeqAction.value.organizationId;
+            const roleId :UUID = storedSeqAction.value.roleId;
+            return state
+              .setIn(['orgs', organizationId, 'grants', roleId], grant.toImmutable())
```
bergieboy: should `grants` be a const?

src/containers/orgs/OrgsSagas.js
```diff
@@ -53,21 +58,67 @@ const { PermissionTypes, PrincipalTypes } = Types;
...
   getOrganizationIntegrationAccount,
   getOrganizationMembers,
+  removeConnections,
 } = OrganizationsApiActions;
 const {
+  addConnectionsWorker,
   getAllOrganizationsWorker,
   getOrganizationWorker,
   getOrganizationIntegrationAccountWorker,
+  removeConnectionsWorker,
```
bergieboy: What are these foreign beasts?

src/containers/orgs/components/OrgConnectionsSection.js
```diff
@@ -0,0 +1,204 @@
...
+        </div>
+      </SectionGrid>
+    );
+  }
+}
+
+const mapStateToProps = (state :Map) => ({
+  requestStates: {
+    [ADD_CONNECTION]: state.getIn(['orgs', ADD_CONNECTION, 'requestState']),
+    [REMOVE_CONNECTION]: state.getIn(['orgs', REMOVE_CONNECTION, 'requestState']),
```
bergieboy: I love this

src/containers/orgs/OrgsSagas.js
UnsungHero97: Whaaat?

src/containers/orgs/OrgsReducer.js
UnsungHero97: I know you all do something different in other apps but I don't know how I feel about it. 

I agree. I don't like having these magic strings all over the place. However, I don't feel like this is any better:

```
import { REDUX_CONSTANTS } from '../../core/redux';
const { ORGS, CONNECTIONS } = REDUX_CONSTANTS;
...
const mapStateToProps = (state :Map) => ({
  requestStates: {
    [UPDATE_ROLE_GRANT]: state.getIn([ORGS, UPDATE_ROLE_GRANT, REQUEST_STATE]),
  }
});
```

bergieboy: Yeah, I think the the benefit is that if you need to change that string for whatever reason, you can do it in one place. You could argue that that the const name could just as easily be misspelled as the string, but the at least linter would yell at you for that. You do you though! Just a suggestion.

UnsungHero97: Solid points. I'll start playing with this approach going forward to see how it feels.

PR #310, opened 2019-12-03
a few subtle fixes with the add / remove modals

UnsungHero97:


src/containers/orgs/components/OrgRolesSection.js
```diff
@@ -136,22 +148,38 @@ class OrgRolesSection extends Component<Props, State> {
...
+        // clicking "close" after having clicked "confirm", i.e. closing the modal after taking action
+        this.setState({
+          isValidRole: true,
+          isVisibleAddRoleModal: false,
+          selectedRoleId: undefined,
+          valueOfRoleTitle: '',
+        });
+        setTimeout(() => {
+          actions.resetRequestState(ADD_ROLE_TO_ORGANIZATION);
+        }, 1000);
```
solomontang: Can you explain the timeout? If you want to guarantee this occurs after the setState(), you can provide a callback to setState as a second param.

```this.setState({...}, () => actions.resetRequestState(...));```

UnsungHero97: I forgot to copy over the timeout comment from below; updated. The `setState()` callback happens too quickly, i.e. the modal transition animation doesn't have enough time to finish.

solomontang: Oh I see. Maybe we should expose the callbacks on the underlying transition itself? Something like this might be defeated if someone reopens the same modal before the timeout occurs right?
https://reactcommunity.org/react-transition-group/css-transition#CSSTransition-prop-onExited

PR #665, opened 2020-10-16
permissions toggle for all properties

UnsungHero97:
toggle is on when every property has permission

Screen Shot 2020-10-16 at 12 27 16 AM

toggle is off if a property does not have permission

Screen Shot 2020-10-16 at 12 27 34 AM


giuliacampana (dismissed):

src/containers/org/components/PermissionsPanel.js
```diff
@@ -200,7 +246,18 @@ const PermissionsPanel = ({
         {`These are the properties that are assigned the ${_lowerCase(permissionType)} permission.`}
       </Typography>
       <Divider isVisible={false} margin={24} />
-      <PropertyTypesCard>
+      <PermissionCard>
+        <PermissionCardSegment>
+          <Typography variant="body1">All Properties</Typography>
+          <IconButton onClick={togglePermissions}>
```
bergieboy, line 252: please add the `aria-label` attribute to buttons that don't contain text. The label should clearly describe the destination, purpose, function, or action for screen reader users.

bergieboy (changes requested):

src/containers/org/components/PermissionsPanel.js
solomontang: Doesn't seem resolved yet

src/containers/org/components/PermissionsPanel.js
```diff
@@ -185,6 +196,41 @@ const PermissionsPanel = ({
...
+          const localAce :?Ace = localPermissions.get(key);
+          const updatedAcePermissions = Set(localAce?.permissions).add(permissionType);
+          const updatedAce = (new AceBuilder())
+            .setPermissions(updatedAcePermissions)
+            .setPrincipal(principal)
+            .build();
+          mutableMap.set(key, updatedAce);
+        });
+      });
+      setLocalPermissions(updatedPermissions);
```
solomontang, line 230: Are there any potential issues if one of these requests fails and a user tries again? What happens if a user ends up granting permissions for something that already exists or removing what doesn't exist

UnsungHero97: `setLocalPermissions` is updating local state, i.e. `useState()`. Persisting the changes happens when they click the save button.

I have not yet added error handling for when permissions are being saved. In theory, if the request fails, nothing in the system has changed... redux store is the same, component state is the same... so I think clicking the save button again should be safe.

I'm not sure how the backend responds to the scenarios you mentioned. I am using the `SET` action type, not the `ADD` / `REMOVE` action type, so I believe the backend will just take what we pass it and overwrite what currently exists.

bergieboy (approved):

solomontang (approved):

PR #686, opened 2020-10-22
add a checkbox for the data set permission

UnsungHero97:
Screen Shot 2020-10-21 at 5 22 37 PM

### DIFF
--- a/src/components/other/Divider.js
+++ b/src/components/other/Divider.js
@@ -30,7 +30,7 @@ const getComputedStyles = ({ isVisible = true, margin } :Props) => {
   `;
 };
 
-const Divider = styled.div`
+const Divider = styled.hr`
   border-color: ${NEUTRAL.N100};
   border-width: 1px;
   ${getComputedStyles}
--- a/src/components/other/Divider.js
+++ b/src/components/other/Divider.js
@@ -30,7 +30,7 @@ const getComputedStyles = ({ isVisible = true, margin } :Props) => {
   `;
 };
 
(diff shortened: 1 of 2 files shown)


src/components/other/Divider.js
```diff
@@ -30,7 +30,7 @@ const getComputedStyles = ({ isVisible = true, margin } :Props) => {
   `;
 };
 
-const Divider = styled.hr`
+const Divider = styled.div`
```
solomontang, line 33: isn't `hr` more semantically appropriate than `div`? in any rare situations where you might need to change the underlying component, you could always use the `as` prop. but honestly the only situation where i could see this happening is if you're using Dividers in between list elements (html5 spec says all elements in a list should be li, including dividers).

src/containers/org/components/PermissionsPanel.js
```diff
@@ -282,12 +300,12 @@ const PermissionsPanel = ({
...
         </IconButton>
       </PanelHeader>
-      <Divider isVisible={false} margin={12} />
-      <Typography variant="body1">
-        {`These are the properties that are assigned the ${_lowerCase(permissionType)} permission.`}
-      </Typography>
       <Divider isVisible={false} margin={24} />
       <PermissionCard>
+        <PermissionCardSegment>
+          <Typography variant="body1">Data Set</Typography>
```
solomontang, line 306: body1 is the default variant

solomontang (approved):

giuliacampana (approved):

PR #702, opened 2020-10-31
data sets landing page + search

UnsungHero97:
Screen Shot 2020-10-31 at 12 01 30 PM

Screen Shot 2020-10-31 at 2 13 09 PM

Screen Shot 2020-11-01 at 3 19 59 PM


src/containers/org/OrgDataSetContainer.js
```diff
@@ -0,0 +1,165 @@
...
+  organizationId,
+} :{|
+  dataSetId :UUID;
+  organizationId :UUID;
+|}) => {
+
+  const dispatch = useDispatch();
+
+  // BUG: this is a bug because getOrSelectDataSets is called in the background
+  const getOrSelectDataSetRS :?RequestState = useRequestState([EDM, GET_OR_SELECT_DATA_SET]);
```
solomontang, line 52: seems a little unclear what part of this is the bug?

src/core/search/sagas/searchOrganizationDataSets.js
```diff
@@ -0,0 +1,116 @@
...
+        searchDataWorker,
+        searchData({
+          ...action.value,
+          entitySetId: SR_DS_META_ESID,
+        })
+      );
+      if (response.error) throw response.error;
+
+      const entitySetIdsHits = response.data.hits
+        // .filter((hit) => getPropertyValue(hit, [FQNS.OL_STANDARDIZED, 0]) === true)
```
solomontang, line 64: Why is this commented out? Even if the property doesn't exist on the set, it doesn't seem like it should affect anything. Same with the similar comment for atlasDataSetIdsHits

solomontang (approved):

src/containers/org/OrgDataSetContainer.js
UnsungHero97: In an earlier version, I was calling `getOrSelectDataSets` (plural), which was the problem. I've since changed it to `getOrSelectDataSet` (singular) and I guess I forgot to remove this comment.

src/core/search/sagas/searchOrganizationDataSets.js
UnsungHero97: Fixed.

solomontang (dismissed):

solomontang (approved):

giuliacampana (approved):

PR #719, opened 2020-11-03
Feature/apps 2507 assign roles

solomontang:
image

TODO: disable members from select when the selected role is already assigned

### DIFF
--- a/src/containers/org/components/AssignRoleModalBody.js
+++ b/src/containers/org/components/AssignRoleModalBody.js
@@ -1,0 +1,51 @@
+// @flow
+import React, { useMemo } from 'react';
+
+import { List } from 'immutable';
+import { AuthUtils } from 'lattice-auth';
+import { Select, Typography } from 'lattice-ui-kit';
+import { PersonUtils } from 'lattice-utils';
+
+import { ModalBody } from '../../../components';
+import { getUserProfileLabel } from '../../../utils/PersonUtils';
+import type { ReactSelectOption } from '../../../types';
+
+const { getUserId } = PersonUtils;
+
+type Props = {
+  onChange :(value :any) => void;
+  members :List;
+}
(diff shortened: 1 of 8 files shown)


src/containers/org/components/AssignRoleModalBody.js
```diff
@@ -0,0 +1,51 @@
...
+  };
+
+  return (
+    <ModalBody>
+      <Typography>Enter the username or email address of the member you wish to assign the role to.</Typography>
+      <Select
+          inputId="assign-to-member"
+          options={memberOptions}
+          placeholder="Search for member"
+          onChange={handleChange} />
```
anzioka: Nit: sort these

UnsungHero97 (approved):

src/containers/org/components/AssignRoleModalBody.js
```diff
@@ -0,0 +1,51 @@
...
+    const options = [];
+    members.forEach((member) => {
+      options.push({
+        label: getUserProfileLabel(member, thisUserId),
+        value: getUserId(member)
+      });
+    });
+    return options;
+  }, [members, thisUserId]);
+
```
anzioka, line 34: this should probably be members.hashCode() since `members` will be different on every render.

anzioka (approved):

UnsungHero97 (approved):

giuliacampana (approved):

src/containers/org/components/AssignRoleModalBody.js
solomontang: it will not be. members is an immutable list of all members in the organization

anzioka: Yes that's true, but I think useMemo will treat it as a new object (I'm probably misremembering)

PR #723, opened 2020-11-04
Surface roles on member page

solomontang:
- Show roles on member page (removable by org owners only)
image


src/containers/org/MemberRoleCard.js
```diff
@@ -0,0 +1,87 @@
...
+  align-items: center;
+  border-radius: 4px;
+  display: grid;
+  grid-gap: 16px;
+  grid-template-columns: 1fr auto;
+  justify-content: space-between;
+  background-color: ${NEUTRAL.N50};
+  border: 1px solid ${NEUTRAL.N50};
+  padding: 8px 24px;
+`;
```
bergieboy, line 26: sort

src/containers/org/MemberRoleCard.js
```diff
@@ -0,0 +1,87 @@
...
+        <StyledRoleIcon />
+        <StyledLink to={rolePath}>
+          {role.title}
+        </StyledLink>
+      </Flex>
+      {
+        unassignable && (
+          <IconButton onClick={handleClick}>
+            <FontAwesomeIcon fixedWidth icon={faTimes} />
+          </IconButton>
```
bergieboy, line 80: aria-label

src/containers/org/members/MembersSection.js
```diff
@@ -145,34 +145,31 @@ const MembersSection = ({
...
-                startIcon={PlusIcon}>
-              Assign Role
-            </Button>
-          )
-        }
+        <Button
+            disabled={disableAddMember}
+            color="primary"
+            onClick={handleAddMemberOnClick}
+            startIcon={PlusIcon}>
```
bergieboy, line 170: sort

anzioka (approved):

bergieboy (approved):

anzioka (dismissed):

src/containers/org/MemberRoleCard.js
```diff
@@ -0,0 +1,87 @@
...
+  };
+
+  return (
+    <RoleCard>
+      <Flex>
+        <StyledRoleIcon />
+        <StyledLink to={rolePath}>
+          {role.title}
+        </StyledLink>
+      </Flex>
```
UnsungHero97, line 75: No change required, just some thoughts. This type of component shows up in several places, and I have a not-really-reusable version of this implemented as `src/components/headers/DataSetTitle.js`. Perhaps there's an abstraction we can make that would satisfy both this case as well as the `DataSetTitle` case.

src/containers/org/MemberRoleCard.js
```diff
@@ -0,0 +1,87 @@
...
+  align-items: center;
+  background-color: ${NEUTRAL.N50};
+  border-radius: 4px;
+  border: 1px solid ${NEUTRAL.N50};
+  display: grid;
+  grid-gap: 16px;
+  grid-template-columns: 1fr auto;
+  justify-content: space-between;
+  padding: 8px 24px;
+`;
```
UnsungHero97, line 26: No change required, just some thoughts. A copy of this exists in `DataSetPermissionsCard`. Perhaps we can consolidate?

src/core/redux/selectors/selectCurrentUserOrgOwner.js
```diff
@@ -0,0 +1,9 @@
+// @flow
+import type { Map } from 'immutable';
+import type { UUID } from 'lattice';
+
+import { IS_OWNER, ORGANIZATIONS } from '../constants';
+
+export default function selectCurrentUserOrgOwner(organizationId :UUID) {
+  return (state :Map) => state.getIn([ORGANIZATIONS, IS_OWNER, organizationId]);
```
UnsungHero97, line 8: The selector returns a boolean:

```
return (state :Map) :boolean => ...
```

src/containers/org/components/RemoveRoleFromMemberModal.js
```diff
@@ -26,26 +26,28 @@ type Props = {
   member :any;
   onClose :() => void;
   organizationId :UUID;
-  roleId :UUID;
+  role ?:Role;
```
UnsungHero97, line 29: Minor flow thing. The `role` prop is required, but your syntax says it's optional. Please switch to `:?Role`, which says "this prop is required by could be `null` or `undefined`. With this change, you can also get rid of the `defaultProps` below.

src/containers/org/members/MembersSection.js
```diff
@@ -145,34 +145,31 @@ const MembersSection = ({
...
 
+  const disableAddMember = selectedRole ? false : !isOwner;
+
+  const handleAddMemberOnClick = () => {
+    if (selectedRole) {
+      setIsVisibleAssignRoleModal(true);
+    }
+    else {
+      setIsVisibleAddMemberToOrgModal(true);
+    }
```
UnsungHero97, line 156: Just to be super defensive, I would explicitly set the visibility of the other modal to false.

UnsungHero97 (changes requested):

src/containers/org/MemberRoleCard.js
```diff
@@ -0,0 +1,87 @@
...
+
+const StyledRoleIcon = styled(RoleIcon)`
+  margin-right: 16px;
+`;
+
+type Props = {
+  onClick :(role :Role) => void;
+  organizationId :UUID;
+  role :Role;
+  unassignable :boolean;
```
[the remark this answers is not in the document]
solomontang: The prop is describing if the role can be unassigned, not if it can be assigned. isUnassignable would be more appropriate

src/containers/org/members/MembersSection.js
solomontang: unless i set up a reducer to do both setStates at once, there is a chance it would incur extra re-renders due to multiple updates. The appropriate modal is completely unmounted when selectedRole is falsy anyway. Leaving as is until there's a tangible reason to change.

UnsungHero97 (approved):

anzioka (approved):

bergieboy (approved):

PR #724, opened 2020-11-07
refactor the modal

UnsungHero97:


src/components/modals/ModalBody.js
```diff
@@ -2,16 +2,42 @@
...
+  children :any;
+  onCleanUp ?:() => void;
+  width ?:number;
+|}) => {
+
+  useEffect(() => () => {
+    if (_isFunction(onCleanUp)) {
+      onCleanUp();
+    }
+  }, [onCleanUp]);
```
solomontang, line 29: I don't think this abstraction is necessary and is actually detrimental. Reasons will be detailed in later comments.

src/containers/org/components/assign-permissions-to-data-set-modal/AssignPermissionsToDataSetModal.js
```diff
@@ -0,0 +1,173 @@
...
+
+const ModalHeader = styled(LUKModalHeader)`
+  padding: 30px 0;
+`;
+
+const ModalFooter = styled(LUKModalFooter)`
+  padding: 30px 0;
+`;
+
+const AssignPermissionsToDataSetModal = ({
```
solomontang, line 37: This is component is not a modal. recommend renaming.

src/containers/org/components/assign-permissions-to-data-set-modal/AssignPermissionsToDataSetModal.js
```diff
@@ -0,0 +1,173 @@
...
+  const [targetDataSetId, setTargetDataSetId] = useState('');
+  const [targetDataSetTitle, setTargetDataSetTitle] = useState('');
+  const [targetPermissionOptions, setTargetPermissionOptions] = useState([]);
+
+  const assignPermissionsToDataSetRS :?RequestState = useRequestState([PERMISSIONS, ASSIGN_PERMISSIONS_TO_DATA_SET]);
+
+  const onCleanUp = useCallback(() => {
+    dispatch(clearSearchState(SEARCH_DATA_SETS_TO_ASSIGN_PERMISSIONS));
+    dispatch(resetRequestState([ASSIGN_PERMISSIONS_TO_DATA_SET]));
+  }, [dispatch]);
```
solomontang, line 65: This can be replaced by calling useEffect(...) here instead of passing a useCallback to some other component.

src/containers/org/components/assign-permissions-to-data-set-modal/AssignPermissionsToDataSetModal.js
```diff
@@ -0,0 +1,173 @@
...
+        withProperties: assignPermissionsToAllProperties,
+      })
+    );
+  };
+
+  if (step === 0) {
+    return (
+      <>
+        <ModalHeader onClickClose={onClose} textTitle="Assign Permissions To Data Set" />
+        <ModalBody onCleanUp={onCleanUp}>
```
solomontang, line 82: Judging from the actions dispatched inside onCleanup, it seems the intention is to clean up when this whole AssignPermissionsToDataSetComponent is unmounted, rather than on each step. Based on the way this is currently set up, I would expect onCleanUp to trigger on each step. 

Not every instance of ModalBody needs to clean up after itself and it feels like you're a little too eager to justify the abstraction of the clean up effect you've created earlier.

In situations where `StepSelectDataSet` does need to clean up after itself, I would still say relocating clean up to be outside of the component isn't necessary. Making abstractions for native hook behavior is dangerous as it obfuscates when and why you might need them. 

It didn't even really save you any lines of code either as the cleanup function is wrapped in a useCallback that would otherwise not be needed

solomontang (changes requested):

src/components/modals/ModalBody.js
UnsungHero97: I think this is worth keeping around.

src/containers/org/components/assign-permissions-to-data-set-modal/AssignPermissionsToDataSetModal.js
UnsungHero97: Yes, the intention is to clean up after the workflow is complete, not in between steps. Strangely, it seems to behave correctly as it was implemented, but that's definitely not what I intended. I've switched to `useEffect`.

solomontang (approved):

anzioka (approved):

PR #766, opened 2020-12-03
Feature/apps 2541 promote staging

solomontang:
# Warning
- ~requires update to lattice-js https://github.com/openlattice/lattice-js/pull/1092~

# Features
- Add menu button to data set view
- Atlas data sets that are not using the `openlattice` schema will have `Promote Table` enabled. Promoted data sets will show a disabled `Promoted` menu item instead.
- The Promoted/Promote Table menu item will be disabled for users who are not owners of the organization or if the data set is not an atlas data set

image
image
image


src/components/headers/ActionWrapper.js
```diff
@@ -0,0 +1,9 @@
+import styled from 'styled-components';
+
+const ActionWrapper = styled.div`
```
UnsungHero97, line 3: How about like `SpaceBetweenWrapper`?

src/components/headers/ActionWrapper.js
```diff
@@ -0,0 +1,9 @@
...
+  display: flex;
+  justify-content: space-between;
+  align-items: flex-start;
```
UnsungHero97, line 6: Please sort.

src/containers/org/components/dataset/DataSetActionButton.js
```diff
@@ -0,0 +1,151 @@
...
+          open={state.menuOpen}
+          transformOrigin={{
+            horizontal: 'right',
+            vertical: 'top',
+          }}>
+        <MenuItem disabled={!isOwner || !isAtlas || isPromoted} onClick={handleOpenPromote}>
+          {
+            isPromoted
+              ? 'Promoted'
+              : 'Promote Table'
```
UnsungHero97, line 133: Please use `Data Set` instead of `Table`.

src/containers/org/components/dataset/PromoteBody.js
```diff
@@ -0,0 +1,72 @@
...
+      tableName,
+    }));
+  };
+
+  return (
+    <StyledBody>
+      <Typography
+          color="textSecondary"
+          gutterBottom>
+        {`Promote the ${tableName} table to the OpenLattice schema.`}
```
UnsungHero97, line 58: Please use `data set` instead of `table`.

UnsungHero97 (changes requested):

src/components/headers/ActionWrapper.js
solomontang: I feel alphabetically sorting CSS properties doesn't always make sense esp since their cascading nature often means some properties depend on/override others. `align-items: flex-start` means nothing without `display: flex`

UnsungHero97: Thanks for the detailed response. That totally makes sense. We've historically alphabetized CSS so that's why I'm pointing it out here. It's not a hard requirement, and of course, do the correct thing where order actually matters.

UnsungHero97 (approved):

anzioka (approved):

giuliacampana (approved):

PR #772, opened 2020-12-07
APPS-2519 remove role from org

bergieboy:
Added 'Remove Role From Organization'  to `RoleActionMenu`.


src/containers/org/roles/RemoveRoleFromOrgModal.js
```diff
@@ -0,0 +1,87 @@
...
+
+  useEffect(() => {
+    if (requestState === RequestStates.SUCCESS) {
+      onClose();
+      // the timeout avoids rendering the modal with new state before the transition animation finishes
+      setTimeout(() => {
+        dispatch(resetRequestState([REMOVE_ROLE_FROM_ORGANIZATION]));
+      }, 1000);
+    }
+  }, [dispatch, requestState, rolePath, onClose]);
```
solomontang, line 73: Remove this effect. If we're going to have the modal close itself on success, we don't need to be using the action modal since those acknowledgements won't be visible anyway. Also, with the above mentioned change about using ResetOnUnmount, the setTimeout can be avoided.

src/containers/org/roles/RemoveRoleFromOrgModal.js
```diff
@@ -0,0 +1,87 @@
...
+  };
+
+  const handleOnClickPrimary = () => {
+    dispatch(
+      removeRoleFromOrganization({
+        organizationId,
+        roleId,
+      })
+    );
+    dispatch(goToRoute(rolePath));
```
solomontang, line 62: This routing looks like it happens regardless of whether or not removeRoleFromOrganization succeeds or fails. Also, this routing seems like it would prevent the reset from happening since the following effect might no longer have a chance to happen.

solomontang (changes requested):

src/containers/org/roles/RemoveRoleFromOrgModal.js
bergieboy: yeah, I posted on slack about this. I think the `RoleActionMenu` belongs in `MemberSection` of the `OrgMembersContainer` for this reason. I don't think you should be able to delete something on role details page.

src/containers/org/roles/RemoveRoleFromOrgModal.js
```diff
@@ -0,0 +1,54 @@
...
+const resetStatePath = [REMOVE_ROLE_FROM_ORGANIZATION];
+
+type Props = {
+  isVisible :boolean;
+  onClose :() => void;
+  organization :Organization;
+  role :Role;
+};
+
+const RoleDetailsModal = ({
```
UnsungHero97, line 21: Copy/pasta? This should be `RemoveRoleFromOrgModal`.

src/containers/org/roles/RemoveRoleFromOrgModal.js
```diff
@@ -0,0 +1,54 @@
...
+        onClose={onClose}
+        textTitle="Delete Role">
+      <ResetOnUnmount
+          path={resetStatePath}
+          message="Are you sure you want to delete this role? This action cannot be undone." />
+    </Modal>
+  );
+};
+
+export default RoleDetailsModal;
```
UnsungHero97, line 54: `RemoveRoleFromOrgModal`

src/containers/org/roles/RemoveRoleFromOrgModal.js
```diff
@@ -0,0 +1,54 @@
...
+
+const RoleDetailsModal = ({
+  isVisible,
+  onClose,
+  organization,
+  role
+} :Props) => {
+  const dispatch = useDispatch();
+  const roleId :UUID = role?.id || '';
+  const organizationId :string = organization?.id || '';
```
UnsungHero97, line 29: The modal doesn't render anything specific about the org or role, so it's better to just pass in `organizationId` and `roleId` as props.

src/containers/org/roles/OrgRoleContainer.js
```diff
@@ -51,18 +53,23 @@ const OrgRoleContainer = ({
...
-
   const orgPath = useMemo(() => (
     Routes.ORG.replace(Routes.ORG_ID_PARAM, organizationId)
   ), [organizationId]);
 
+  useEffect(() => {
+    if (organization && !role) {
+      const membersPath = Routes.ORG_MEMBERS.replace(Routes.ORG_ID_PARAM, organizationId);
+      dispatch(goToRoute(membersPath));
+    }
```
UnsungHero97, line 70: I think the condition should be if `REMOVE_ROLE_FROM_ORGANIZATION` is success, then route. It's possible for the id in the url to be incorrect, in which case `role` would not be defined, in which case it's probably better to show an error message.

UnsungHero97 (changes requested):

src/containers/org/roles/OrgRoleContainer.js
bergieboy: So should I not redirect and instead show an error message in the `OrgRoleContainer`? This flow is just weird with the deleting the role from this component.

bergieboy: Then I could keep the modal open with a success or a fail?

bergieboy: Once the role is deleted, the modal will close, and you will be rerouted to the members view.

UnsungHero97 (approved):

solomontang (approved):

PR #774, opened 2020-12-11
org object permissions page

UnsungHero97:
`ObjectPermissionsCardStack` and `ObjectPermissionsCard` are meant to be generic (though they are not fully at the moment) with the hope that we can reuse them for other permissions pages.

Screen Shot 2020-12-11 at 9 47 37 AM
Screen Shot 2020-12-11 at 9 49 54 AM
Screen Shot 2020-12-11 at 9 50 19 AM


src/containers/org/OrgObjectPermissionsContainer.js
```diff
@@ -0,0 +1,151 @@
...
+      if (!options) {
+        setFilterByPermissionTypes([]);
+      }
+      else {
+        setFilterByPermissionTypes(options.map((option) => option.value));
+      }
+    };
+
+    const handleOnChangeFilterQuery = (event :SyntheticInputEvent<HTMLInputElement>) => {
+      setFilterByQuery(event.target.value || '');
```
solomontang, line 102: consider debouncing this as updating this value every keystroke will probably cause jank

src/containers/org/OrgObjectPermissionsContainer.js
```diff
@@ -0,0 +1,151 @@
...
+import type { ReactSelectOption } from '../../types';
+
+const { PermissionTypes } = Types;
+
+const PERMISSION_TYPE_OPTIONS = [
+  { label: PermissionTypes.OWNER.toLowerCase(), value: PermissionTypes.OWNER },
+  { label: PermissionTypes.READ.toLowerCase(), value: PermissionTypes.READ },
+  { label: PermissionTypes.WRITE.toLowerCase(), value: PermissionTypes.WRITE },
+  { label: PermissionTypes.LINK.toLowerCase(), value: PermissionTypes.LINK },
+  { label: PermissionTypes.MATERIALIZE.toLowerCase(), value: PermissionTypes.MATERIALIZE },
```
solomontang, line 52: Isn't there also discover?

src/core/permissions/sagas/updatePermissions.js
```diff
@@ -0,0 +1,104 @@
...
+      //   ));
+      //   const targetAce :Ace = storedAces.get(targetIndex);
+      //   const updatedPermissions :Set<PermissionType> = Set(targetAce.permissions).subtract(ace.permissions);
+      //   if (updatedPermissions.isEmpty() || updatedPermissions.equals(Set([PermissionTypes.DISCOVER]))) {
+      //     finalActionType = ActionTypes.SET;
+      //     finalAces = storedAces.delete(targetIndex);
+      //     // any aces with empty permissions will cause the request will fail, so we'll filter those out too
+      //     finalAces = finalAces.filter((storedAce :Ace) => storedAce.permissions.length > 0);
+      //   }
+      // }
```
solomontang, line 67: a separate branch seems like an appropriate place to keep this

solomontang (changes requested):

src/containers/org/OrgObjectPermissionsContainer.js
```diff
@@ -0,0 +1,151 @@
...
+
+  const dispatch = useDispatch();
+  const [filterByPermissionTypes, setFilterByPermissionTypes] = useState([]);
+  const [filterByQuery, setFilterByQuery] = useState('');
+
+  const getOrgObjectPermissionsRS :?RequestState = useRequestState([PERMISSIONS, GET_ORGANIZATION_OBJECT_PERMISSIONS]);
+
+  const organization :?Organization = useSelector(selectOrganization(organizationId));
+
+  const key = useMemo(() => List([organizationId]), [organizationId]);
```
anzioka, line 73: wondering if this optimization is necessary since this doesn't seem to be an expensive computation

src/containers/org/components/PermissionsPanel.js
```diff
@@ -289,36 +290,42 @@ const PermissionsPanel = ({
...
+            <Typography>All properties</Typography>
+            <IconButton
+                aria-label="permissions toggle for all properties"
+                onClick={togglePermissionAssignmentAll}>
+              <FontAwesomeIcon
+                  color={isPermissionAssignedToAll ? PURPLE.P300 : NEUTRAL.N500}
+                  fixedWidth
+                  icon={faToggleOn}
+                  transform={{ rotate: isPermissionAssignedToAll ? 0 : 180 }}
+                  size="lg" />
```
anzioka, line 310: please sort

src/containers/org/components/PermissionsPanel.js
```diff
@@ -289,36 +290,42 @@ const PermissionsPanel = ({
...
+            <Typography>Only non-pii properties</Typography>
+            <IconButton
+                aria-label="permissions toggle for only non-pii properties"
+                onClick={togglePermissionAssignmentOnlyNonPII}>
+              <FontAwesomeIcon
+                  color={isPermissionAssignedToOnlyNonPII ? PURPLE.P300 : NEUTRAL.N500}
+                  fixedWidth
+                  icon={faToggleOn}
+                  transform={{ rotate: isPermissionAssignedToOnlyNonPII ? 0 : 180 }}
+                  size="lg" />
```
anzioka, line 325: please sort

src/containers/permissions/ObjectPermissionsCard.js
```diff
@@ -0,0 +1,157 @@
...
+  // TODO: spinner doesn't have a min-height like the checkbox, so the row height drops when showing the spinner
+  return (
+    <CardSegment padding="24px 0">
+      <SpaceBetweenGrid>
+        <Typography component="span">{title}</Typography>
+        <SpaceBetweenGrid gap={8}>
+          <Typography component="span">{permissions}</Typography>
+          <IconButton onClick={() => setIsOpen(!isOpen)}>
+            <FontAwesomeIcon fixedWidth icon={isOpen ? faChevronUp : faChevronDown} />
+          </IconButton>
```
anzioka, line 126: do you need `aria-label` for this button?

src/containers/permissions/ObjectPermissionsCardStack.js
```diff
@@ -0,0 +1,93 @@
...
+  const organizationMembers :Map<Principal, Map> = useMemo(() => (
+    Map().withMutations((mutableMap :Map<string, Map>) => {
+      members.forEach((member :Map) => {
+        const principal :?Principal = getPrincipal(member);
+        if (principal) {
+          mutableMap.set(principal, member);
+        }
+      });
+    })
+  ), [members]);
```
anzioka, line 58: I'm curious how useMemo and other hooks compare equality between immutable collection.

src/containers/org/OrgObjectPermissionsContainer.js
UnsungHero97: We're ignoring discover:

https://jira.openlattice.com/browse/LATTICE-2578
https://jira.openlattice.com/browse/APPS-2381

src/containers/permissions/ObjectPermissionsCardStack.js
UnsungHero97: React uses `Object.is`:

https://reactjs.org/docs/hooks-reference.html#bailing-out-of-a-state-update
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is#Description

Here, the `members` object should be the same across renders because it's not being "computed". It comes from the redux store, so we expect to get the same object every time.

src/containers/org/OrgObjectPermissionsContainer.js
UnsungHero97: I wonder this as well... yea, I'll get rid of it since it's unlikely to be much of an optimization. Plus, the docs say:

> You may rely on useMemo as a performance optimization, not as a semantic guarantee. In the future, React may choose to “forget” some previously memoized values and recalculate them on next render, e.g. to free memory for offscreen components. Write your code so that it still works without useMemo — and then add it to optimize performance.

anzioka (approved):

solomontang (approved):

solomontang (approved):

PR #795, opened 2020-12-22
data set object permissions page

UnsungHero97:
Screen Shot 2020-12-22 at 9 53 56 AM
Screen Shot 2020-12-22 at 9 54 17 AM
Screen Shot 2020-12-22 at 9 55 18 AM

PR #796, opened 2020-12-23
permissions ui updates + some refactoring + cleanup

UnsungHero97:
https://github.com/openlattice/lattice-orgs/pull/795


src/containers/org/members/OrgMemberContainer.js
```diff
@@ -85,53 +92,83 @@ const OrgMemberContainer = ({
...
+                <AssignPermissionsToDataSet
+                    onClose={() => setIsVisibleAddDataSetModal(false)}
+                    organizationId={organizationId}
+                    principal={memberPrincipal}
+                    step={step}
+                    stepBack={stepBack}
+                    stepNext={stepNext} />
+              )
+            }
+          </StepsController>
```
solomontang, line 169: StepsController should probably use a render prop instead of intercepting `children` as it is unclear which scope step, stepBack, stepNext come from. also would this still work if there are multiple children? if not, we should try not to break typically sanctioned jsx patterns.

also it seems AssignPermissionsToDataSet is not designed to be rendered without the steps. Consider moving the StepsController to be contained inside the component.

src/containers/permissions/DataSetPermissionsContainer.js
```diff
@@ -0,0 +1,212 @@
...
+  const initializePermissionsRS :?RequestState = useRequestState([PERMISSIONS, INITIALIZE_DATA_SET_PERMISSIONS]);
+
+  const atlasDataSetIds :Set<UUID> = useSelector(selectOrganizationAtlasDataSetIds(organizationId));
+  const atlasDataSets :Map<UUID, Map> = useSelector(selectAtlasDataSets(atlasDataSetIds));
+  const entitySetIds :Set<UUID> = useSelector(selectOrganizationEntitySetIds(organizationId));
+  const entitySets :Map<UUID, EntitySet> = useSelector(selectEntitySets(entitySetIds));
+  const atlasDataSetIdsHash :number = atlasDataSetIds.hashCode();
+  const atlasDataSetsHash :number = atlasDataSets.hashCode();
+  const entitySetIdsHash :number = entitySetIds.hashCode();
+  const entitySetsHash :number = entitySets.hashCode();
```
solomontang, line 84: we could avoid using this hashCode hack if the entitySets were already grouped by organizationId in the redux store.

src/core/redux/selectors/selectIsOwner.js
```diff
@@ -0,0 +1,23 @@
...
+
+import { Map, getIn } from 'immutable';
+import { ValidationUtils } from 'lattice-utils';
+import type { UUID } from 'lattice';
+
+import { IS_OWNER, PERMISSIONS } from '../constants';
+
+const { isValidUUID } = ValidationUtils;
+
+export default function selectIsOwner(id :UUID) {
```
solomontang, line 13: rename to make it clear that this checks permission

solomontang (changes requested):

src/containers/permissions/PermissionsActionsGrid.js
```diff
@@ -20,28 +21,42 @@ const StyledActionsGrid :ComponentType<ActionsGridProps> = styled(ActionsGrid)`
...
+const PermissionsActionsGrid = ({
+  assignPermissionsText,
   onChangeFilterByPermissionTypes,
   onChangeFilterByQuery,
+  onClickAssignPermissions,
 } :{|
+  assignPermissionsText :string;
   onChangeFilterByPermissionTypes :(permissionTypes :PermissionType[]) => void;
   onChangeFilterByQuery :(query :string) => void;
+  onClickAssignPermissions :() => void;
```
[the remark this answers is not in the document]
UnsungHero97: It's required because it's meant to be implemented in https://jira.openlattice.com/browse/APPS-2657 and https://jira.openlattice.com/browse/APPS-2658 and https://jira.openlattice.com/browse/APPS-2678.

src/containers/org/OrgDataSetObjectPermissionsContainer.js
```diff
@@ -74,9 +76,11 @@ const OrgDataSetObjectPermissionsContainer = ({
         <Typography>
           Below are the users, roles, and organizations that are granted permissions on this object.
         </Typography>
-        <ObjectPermissionsActionsGrid
+        <PermissionsActionsGrid
+            assignPermissionsText="Add permission"
             onChangeFilterByPermissionTypes={setFilterByPermissionTypes}
-            onChangeFilterByQuery={setFilterByQuery} />
+            onChangeFilterByQuery={setFilterByQuery}
+            onClickAssignPermissions={() => {}} />
```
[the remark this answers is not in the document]
UnsungHero97: It's intentional as it's meant to be implemented as part of https://jira.openlattice.com/browse/APPS-2657 and https://jira.openlattice.com/browse/APPS-2658 and https://jira.openlattice.com/browse/APPS-2678.

src/containers/org/OrgObjectPermissionsContainer.js
```diff
@@ -67,9 +67,11 @@ const OrgObjectPermissionsContainer = ({
           <Typography>
             Below are the users and roles that are granted permissions on this object.
           </Typography>
-          <ObjectPermissionsActionsGrid
+          <PermissionsActionsGrid
+              assignPermissionsText="Add permission"
               onChangeFilterByPermissionTypes={setFilterByPermissionTypes}
-              onChangeFilterByQuery={setFilterByQuery} />
+              onChangeFilterByQuery={setFilterByQuery}
+              onClickAssignPermissions={() => {}} />
```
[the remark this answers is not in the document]
UnsungHero97: It's coming in https://jira.openlattice.com/browse/APPS-2657 and https://jira.openlattice.com/browse/APPS-2658 and https://jira.openlattice.com/browse/APPS-2678.

src/containers/permissions/DataSetPermissionsContainer.js
UnsungHero97: I realized I don't need to call `hashCode()` on the ids; updated. To update the data sets would be a larger change, better suited as a follow-up pr.

src/containers/org/members/OrgMemberContainer.js
UnsungHero97: I feel like we're going in circles here. Let's talk.

src/containers/org/OrgRoleContainer.js
```diff
@@ -55,6 +61,9 @@ const OrgRoleContainer = ({
   rolesRoute :string;
 |}) => {
 
+  const [filterByPermissionTypes, setFilterByPermissionTypes] = useState([]);
+  const [filterByQuery, setFilterByQuery] = useState('');
```
anzioka, line 65: Perhaps no change required here, but `filterByPermissionTypes` and `filterByQuery` as names of the stateful values sound rather awkward

src/containers/org/OrgRoleContainer.js
```diff
@@ -69,47 +78,81 @@ const OrgRoleContainer = ({
...
+                    onClose={handleOnClosePermissionsPanel}
+                    principal={role.principal}
+                    permissionType={selection.permissionType} />
+              </PanelColumn>
+            )
+          }
+        </AppContentGrid>
+        <Modal
+            isVisible={isVisibleAddDataSetModal}
+            onClose={() => {}}
```
anzioka, line 136: why is this onClose callback empty?

src/utils/getUserProfile.js
```diff
@@ -0,0 +1,36 @@
...
+import { LangUtils } from 'lattice-utils';
+
+import type { UserProfile } from '../types';
+
+const { isNonEmptyString } = LangUtils;
+
+export default function getUserProfile(user :any) :UserProfile {
+
+  const auth0UserProfile = get(user, 'profile', user);
+
```
anzioka, line 15: confused about this: the fallback value is the same collection from which we are trying to read?

src/utils/getUserProfile.js
```diff
@@ -0,0 +1,36 @@
...
+  let givenName = get(auth0UserProfile, 'given_name');
+  if (!isNonEmptyString(givenName)) {
+    givenName = '';
+  }
+
+  let name = get(auth0UserProfile, 'name');
+  if (!isNonEmptyString(name)) {
+    name = '';
+  }
```
anzioka, line 29: Won't providing a fallback value (empty string) for the immutable `get` achieve the same goal in these three instances?

src/containers/org/OrgRoleContainer.js
UnsungHero97: I went with this basically because I wanted them to be together when sorting. I considered `permissionTypesToFilterBy` and `queryToFilterBy`, but those are rather lengthy and won't necessarily be together in props. I also considered `filterQuery` and `filterPermissionTypes`, but I like `filterBy` better.

src/containers/org/OrgRoleContainer.js
UnsungHero97: For this modal, I'm passing `withHeader={false}` and using my own header component, so `onClose` is a noop because it doesn't do anything. Perhaps this should be a flow fix in LUK.

src/utils/getUserProfile.js
UnsungHero97: Correct. `user` could be one of multiple objects. If it has `'profile'`, then it's `com.openlattice.organization.OrganizationMember`. If it doesn't, then it's likely going to be `com.openlattice.directory.pojo.Auth0UserBasic`.

Note: this should be moved to `lattice-utils`.

src/utils/getUserProfile.js
UnsungHero97: Well... yea... I guess I'm just being extra defensive in the case where any of these fields are not a strings, which should not be possible, but 🤷‍♂️

src/containers/org/members/OrgMemberContainer.js
```diff
@@ -85,53 +92,83 @@ const OrgMemberContainer = ({
...
+                    onClose={handleOnClosePermissionsPanel}
+                    principal={memberPrincipal}
+                    permissionType={selection.permissionType} />
+              </PanelColumn>
+            )
+          }
+        </AppContentGrid>
+        <Modal
+            isVisible={isVisibleAddDataSetModal}
+            onClose={() => {}}
```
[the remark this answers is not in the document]
UnsungHero97: For this modal, I'm passing withHeader={false} and using my own header component, so onClose is a noop because it doesn't do anything. Perhaps this should be a flow fix in LUK.

src/containers/org/OrgRoleContainer.js
anzioka: well, `filterByPermissionTypes` with no additional context would most likely be assumed to be the name of a function, not an array as is the case here. The same could be said about `filterByQuery`.

anzioka (approved):

anzioka (approved):

solomontang (approved):

src/containers/permissions/assign-permissions-to-data-set/AssignPermissionsToDataSetModalBody.js
```diff
@@ -0,0 +1,155 @@
...
+                <>
+                  <ModalBody>
+                    <StepConfirm
+                        assignPermissionsToAllProperties={assignPermissionsToAllProperties}
+                        assignPermissionsToDataSetRS={assignPermissionsToDataSetRS}
+                        targetDataSetTitle={targetDataSetTitle}
+                        targetPermissionOptions={targetPermissionOptions} />
+                  </ModalBody>
+                  <ModalFooter
+                      isPendingPrimary={assignPermissionsToDataSetRS === RequestStates.PENDING}
```
bergieboy, line 139: is there a reason we're not using `isPending` from `ReduxUtils` in `lattice-utils`?

src/containers/permissions/DataSetPermissionsContainer.js
```diff
@@ -0,0 +1,210 @@
...
+          <Spinner />
+        )
+      }
+      {
+        reducedRS === RequestStates.SUCCESS && filteredPermissionsCount === 0 && (
+          <Typography align="center">No datasets.</Typography>
+        )
+      }
+      {
+        reducedRS === RequestStates.SUCCESS && filteredPermissionsCount !== 0 && (
```
bergieboy, line 181: is there a reason we aren't using `isPending` and `isSuccess` from `ReduxUtils` in `lattice-utils`?

src/containers/permissions/ObjectPermissionsCard.js
```diff
@@ -91,7 +89,11 @@ const ObjectPermissionsCard = ({
...
 
   const updatePermissionsRS :?RequestState = useRequestState([PERMISSIONS, UPDATE_PERMISSIONS]);
+
   const user :Map = useSelector(selectUser(principal.id));
+  const thisUserInfo = AuthUtils.getUserInfo() || { id: '' };
+  const thisUserId = thisUserInfo.id;
+  const title :string = getPrincipalTitle(principal, user, thisUserId);
 
   useEffect(() => {
     if (updatePermissionsRS !== RequestStates.PENDING) {
```
bergieboy, line 99: same comment about `ReduxUtils`

bergieboy (approved):

PR #838, opened 2021-01-27
APPS-2722: Add Button to Regenerate Atlas Credentials in AccountContainer

bergieboy:
- Adds `regenerateCredentials` to `AccountReducer`
- Adds `regenerateCredentialsWatcher` to sagas
- Removed Spinner during `STANDBY` and `PENDING` request states for `getAtlasCredentials`, and instead disabled all buttons related to atlas credentials during these states and the `PENDING` request state for `regenerateCredential`. 
- Add button to regenerate atlas credentials
Screen Shot 2021-01-26 at 9 23 16 PM
- Updated to show `Snackbar` for `FAILURE` and `SUCCESS` states.


src/containers/account/AccountContainer.js
```diff
@@ -6,28 +6,34 @@ import React, { useEffect } from 'react';
...
 import { AuthUtils } from 'lattice-auth';
-import { PrincipalsApiActions } from 'lattice-sagas';
 import { AppContentWrapper, Input, Typography } from 'lattice-ui-kit';
-import { useRequestState } from 'lattice-utils';
+import { ReduxUtils, useRequestState } from 'lattice-utils';
 import { useDispatch, useSelector } from 'react-redux';
-import { RequestStates } from 'redux-reqseq';
 import type { UserInfo } from 'lattice-auth';
 import type { RequestState } from 'redux-reqseq';
+import { PrincipalsApiActions } from 'lattice-sagas';
```
UnsungHero97, line 14: Please run the import auto sorter.

src/containers/account/AccountContainer.js
```diff
@@ -94,30 +113,30 @@ const AccountContainer = () => {
...
+          <ActionsGrid fit>
+            {PasswordInput}
+            <CopyButton
+                aria-label="copy atlas credential"
+                isDisabled={atlasCredentialsPending}
+                onClick={() => clipboardWriteText(atlasCredentials.get('credential'))} />
+            <RefreshButton
+                aria-label="regenerate atlas credential"
+                isDisabled={atlasCredentialsPending}
+                onClick={regenerateAtlasCredential} />
```
UnsungHero97, line 137: I think you should use the `isLoading` prop to show a spinner when `regenerateAtlasCredentialsRS` is pending.

src/containers/account/AccountContainer.js
```diff
@@ -40,10 +46,31 @@ const AccountContainer = () => {
...
 
   const getAtlasCredentialsRS :?RequestState = useRequestState([ACCOUNT, GET_ATLAS_CREDENTIALS]);
+  const regenerateAtlasCredentialsRS :?RequestState = useRequestState([ACCOUNT, REGENERATE_CREDENTIAL]);
   const atlasCredentials :Map = useSelector(selectAtlasCredentials());
 
+  const getAtlasCredentialsIsPending = isPending(getAtlasCredentialsRS);
+  const getAtlasCredentialsIsStandby = isStandby(getAtlasCredentialsRS);
+
+  const regeneratingAtlasCredentials = isPending(regenerateAtlasCredentialsRS);
+  const atlasCredentialsRegenerated = isSuccess(regenerateAtlasCredentialsRS);
```
UnsungHero97, line 56: I think it's a good idea to handle the failure case. I don't have great ideas on how to design it, but one thought that comes to mind is to use a toast (similar to what we see in jira):

https://material-ui.com/components/snackbars/#customized-snackbars

src/containers/account/AccountContainer.js
```diff
@@ -52,14 +79,6 @@ const AccountContainer = () => {
     dispatch(clearAtlasCredentials());
   }, [dispatch]);
 
-  if (getAtlasCredentialsRS === RequestStates.PENDING || getAtlasCredentialsRS === RequestStates.STANDBY) {
-    return (
-      <AppContentWrapper>
-        <Spinner />
-      </AppContentWrapper>
-    );
-  }
```
UnsungHero97, line 61: Why did you decide to remove this?

src/containers/account/reducers/regenerateCredentialReducer.js
```diff
@@ -0,0 +1,36 @@
...
+
+export default function reducer(state :Map, action :SequenceAction) {
+
+  return regenerateCredential.reducer(state, action, {
+    REQUEST: () => state
+      .setIn([REGENERATE_CREDENTIAL, REQUEST_STATE], RequestStates.PENDING)
+      .setIn([REGENERATE_CREDENTIAL, action.id], action),
+    SUCCESS: () => {
+      const storedAction = state.getIn([REGENERATE_CREDENTIAL, action.id]);
+      if (storedAction) {
```
UnsungHero97, line 22: No change required, just a small note. Since you're not using `storedAction`, you can just do `state.hasIn` like you do in the `FAILURE` block.

UnsungHero97 (changes requested):

src/containers/account/AccountContainer.js
bergieboy: 1000%

src/containers/account/AccountContainer.js
bergieboy: I decided to remove it because when you regenerate the atlas credential, you have to call `getAtlasCredentials`, which makes the whole page really jerky with that spinner. Decided to use these loading states to disable the buttons instead. I like your idea above though. I'll change the prop to `isLoading`.

src/containers/account/AccountContainer.js
```diff
@@ -40,10 +62,65 @@ const AccountContainer = () => {
...
 
   const thisUserInfo :UserInfo = AuthUtils.getUserInfo() || {};
 
+  useEffect(() => {
+    if (regenerateAtlasCredentialsIsSuccess && getAtlasCredentialsIsSuccess) {
+      setSnackbarText(REGENERATE_SUCCESS);
+      setSnackbarSeverity('success');
+      setSnackbarOpen(true);
+    }
+    if (regenerateAtlasCredentialsIsFailure) {
```
UnsungHero97, line 104: This should be an `else if`.

UnsungHero97 (changes requested):

UnsungHero97 (approved):

solomontang (approved):

PR #846, opened 2021-02-04
switch to data set metadata entity

UnsungHero97:


src/core/edm/sagas/getOrgDataSetColumnsFromMeta.js
```diff
@@ -0,0 +1,113 @@
...
+      .map((column :Map) => column.mapKeys((key :string) => FQN.of(key)))
+      .groupBy((column :Map) => getPropertyValue(column, [FQNS.OL_DATA_SET_ID, 0]));
+
+    workerResponse = { data: dataSetsColumns };
+    yield put(getOrgDataSetColumnsFromMeta.success(action.id, workerResponse.data));
+  }
+  catch (error) {
+    workerResponse = { error };
+    LOG.error(action.type, error);
+    yield put(getOrgDataSetColumnsFromMeta.failure(action.id, error));
```
bergieboy, line 96: `toSagaError`?

src/core/edm/sagas/getOrgDataSetsFromMeta.js
```diff
@@ -0,0 +1,89 @@
...
+      .mapKeys((_, dataSet :Map) => getPropertyValue(dataSet, [FQNS.OL_ID, 0]))
+      .map((dataSet :Map) => dataSet.mapKeys((key :string) => FQN.of(key)));
+
+    workerResponse = { data: dataSets };
+    yield put(getOrgDataSetsFromMeta.success(action.id, workerResponse.data));
+  }
+  catch (error) {
+    workerResponse = { error };
+    LOG.error(action.type, error);
+    yield put(getOrgDataSetsFromMeta.failure(action.id, error));
```
bergieboy, line 72: `toSagaError`?

src/core/edm/sagas/initializeOrganizationDataSet.js
```diff
@@ -0,0 +1,186 @@
...
+        .build()
+    ];
+    const authorizationsCall = call(getAuthorizationsWorker, getAuthorizations(accessChecks));
+
+    const dataSetSearchCall = call(
+      searchEntitySetDataWorker,
+      searchEntitySetData({
+        constraints: [{
+          constraints: [{
+            searchTerm: `entity.${propertyTypeIds.get(FQNS.OL_ID)}:${JSON.stringify(dataSetId)}`,
```
bergieboy, line 102: No change, but I've seen this a couple of times. Maybe we add a a `getEntitySetSearchTerm` or a more general `getSearchTerm` to `lattice-utils`? 
```
const getSearchTerm = (propertyTypeId, value) => `entity.${propertyTypeId}:${JSON.stringify(value)}`;
```

src/containers/permissions/DataSetPermissionsContainer.js
```diff
@@ -162,47 +135,37 @@ const DataSetPermissionsContainer = ({
...
-                  isOpen={openDataSetId === dataSetId}
-                  key={dataSetId}
-                  onClick={toggleOpenDataSetCard}
-                  onSelect={onSelect}
-                  organizationId={organizationId}
-                  principal={principal}
-                  selection={selection} />
-            );
-          })
+        isSuccess(getDataSetPermissionsPageRS) && totalPermissions !== 0 && (
```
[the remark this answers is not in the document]
UnsungHero97: Well, `requestIsSuccess` was the result of the reduced request state, which doesn't exist anymore. Separately, regarding your comment about pulling it into a variable, I don't necessary find much value in that. Yes, it avoids the overhead of the function call, but all the function does is an `===` equivalence check, which is fast. Objectively, `isSuccess(getDataSetPermissionsPageRS)` and `getDataSetPermissionsPageIsSuccess` are basically the same amount of characters, so that's a no-op.

src/core/edm/sagas/initializeOrganizationDataSet.js
UnsungHero97: I agree, wrapping this is a good idea. I've thought about it a bit, just not sure how to do it. A generic `getSearchTerm` doesn't make sense because the format matters. `getEntitySetSearchTerm` makes more sense but that also feels a bit off. At the same time, something like `getPropertyTypeSearchTermToMatchExactValue` is pretty ridiculous 😆 .

src/components/typography/DataSetTitle.js
```diff
@@ -5,20 +5,24 @@
...
 
 import { AtlasDataSetIcon, EntitySetIcon } from '../../assets/svg/icons';
-import { getDataSetField, isAtlasDataSet } from '../../utils';
+import { FQNS } from '../../core/edm/constants';
+import { isAtlasDataSet } from '../../utils';
 import { GapGrid } from '../grids';
 
+const { getPropertyValue } = DataUtils;
+
+// TODO: this will have to change because we'll no longer be using the data set objects directly
```
anzioka, line 18: Is this TODO handled in a separate PR?

src/core/permissions/reducers/getDataSetPermissionsPageReducer.js
```diff
@@ -0,0 +1,37 @@
...
+      return state;
+    },
+    FAILURE: () => {
+      if (state.hasIn([GET_DATA_SET_PERMISSIONS_PAGE, action.id])) {
+        return state
+          .set(DATA_SET_PERMISSIONS_PAGE, Map())
+          .setIn([GET_DATA_SET_PERMISSIONS_PAGE, ERROR], action.value)
+          .setIn([GET_DATA_SET_PERMISSIONS_PAGE, REQUEST_STATE], RequestStates.FAILURE);
+      }
+      return state;
```
anzioka, line 33: Do you need an initial state for this reducer? Looks like `GET_DATA_SET_PERMISSIONS_PAGE` should initially have a STANDBY requestState.

src/core/permissions/sagas/getDataSetPermissionsPage.js
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    const propertyTypes :Map<UUID, PropertyType> = yield select(selectPropertyTypes(REQUIRED_PROPERTY_TYPES));
+    const propertyTypeIds :Map<FQN, UUID> = propertyTypes.map((propertyType) => propertyType.type).flip();
+    if (propertyTypeIds.count() !== REQUIRED_PROPERTY_TYPES.length) {
+      throw new Error(ERR_MISSING_PROPERTY_TYPE);
+    }
+
+    if (initialize === true) {
+      // NOTE: we only want to call getOrgDataSetsFromMeta once, not for every page
```
anzioka, line 94: why not just `if(initialize) {...}`

src/core/permissions/sagas/getDataSetPermissionsPage.js
```diff
@@ -0,0 +1,198 @@
...
+    const dataSets :Map<UUID, Map> = yield select(selectOrgDataSets(organizationId));
+    const dataSetKeys :List<List<UUID>> = dataSets.keySeq().map((dataSetId :UUID) => List([dataSetId])).toList();
+
+    if (initialize === true) {
+      // NOTE: we only want to call getPermissions once, not for every page
+      // NOTE: we don't need response.data because the redux store will be populated
+      response = yield call(getPermissionsWorker, getPermissions(dataSetKeys));
+      if (response.error) throw response.error;
+    }
+
```
anzioka, line 109: I think this `if block` and the previous should be coalesced for readability

That is:
```
const dataSets :Map<UUID, Map> = yield select(selectOrgDataSets(organizationId));
const dataSetKeys :List<List<UUID>> = dataSets.keySeq().map((dataSetId :UUID) => List([dataSetId])).toList();

if (initialize) {
  response = yield call(getOrgDataSetsFromMetaWorker, getOrgDataSetsFromMeta({ organizationId }));
  if (response.error) throw response.error;

  response = yield call(getPermissionsWorker, getPermissions(dataSetKeys));
  if (response.error) throw response.error;
}
```

src/containers/permissions/DataSetPermissionsContainer.js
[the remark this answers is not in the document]
bergieboy: yeah, it's a personal preference, so you do you!

src/core/edm/sagas/initializeOrganizationDataSet.js
bergieboy: you could pass it an `exact :boolean` arg that would handle that in the `getEntitySetSearchTerm` fn.

bergieboy (dismissed):

src/core/redux/selectors/selectOrgDataSetsColumns.js
```diff
@@ -0,0 +1,31 @@
...
+export default function selectOrgDataSetsColumns(organizationId :UUID, dataSetIds ?:UUID[]) {
+
+  return (state :Map) :Map<UUID, List<Map<FQN, List>>> => {
+
+    if (isValidUUID(organizationId)) {
+      if (dataSetIds) {
+        return Map().withMutations((mutableMap :Map<UUID, Map>) => {
+          dataSetIds.forEach((dataSetId :UUID) => {
+            const dataSetColumns = getIn(state, [EDM, ORG_DATA_SET_COLUMNS, organizationId, dataSetId]) || List();
+            mutableMap.set(dataSetId, dataSetColumns);
```
anzioka, line 22: same comment about `getIn()` ifNotSetValue

src/core/redux/selectors/selectMyKeys.js
```diff
@@ -0,0 +1,18 @@
...
+  getIn,
+} from 'immutable';
+import type { UUID } from 'lattice';
+
+import { MY_KEYS, PERMISSIONS } from '../constants';
+
+export default function selectMyKeys() {
+
+  return (state :Map) :Set<List<UUID>> => getIn(state, [PERMISSIONS, MY_KEYS]) || Set();
+}
```
anzioka, line 18: no change required, but I prefer passing the `IfNotSetValue` (in this case `Set()`) as the third parameter to `getIn()`

src/core/redux/selectors/selectOrgDataSet.js
```diff
@@ -0,0 +1,23 @@
...
+
+const { isValidUUID } = ValidationUtils;
+
+export default function selectOrgDataSet(organizationId :UUID, dataSetId :UUID) {
+
+  return (state :Map) :Map<FQN, List> => {
+
+    if (isValidUUID(organizationId) && isValidUUID(dataSetId)) {
+      return getIn(state, [EDM, ORG_DATA_SETS, organizationId, dataSetId]) || Map();
+    }
```
anzioka, line 19: same comment about `getIn()` ifNotSetValue

src/core/redux/selectors/selectOrgDataSetColumns.js
```diff
@@ -0,0 +1,23 @@
...
+
+const { isValidUUID } = ValidationUtils;
+
+export default function selectOrgDataSetColumns(organizationId :UUID, dataSetId :UUID) {
+
+  return (state :Map) :List<Map<FQN, List>> => {
+
+    if (isValidUUID(organizationId) && isValidUUID(dataSetId)) {
+      return getIn(state, [EDM, ORG_DATA_SET_COLUMNS, organizationId, dataSetId]) || List();
+    }
```
anzioka, line 19: same comment about `getIn()` ifNotSetValue

src/core/redux/selectors/selectOrgDataSets.js
```diff
@@ -0,0 +1,31 @@
...
+export default function selectOrgDataSets(organizationId :UUID, dataSetIds ?:UUID[]) {
+
+  return (state :Map) :Map<UUID, Map<FQN, List>> => {
+
+    if (isValidUUID(organizationId)) {
+      if (dataSetIds) {
+        return Map().withMutations((mutableMap :Map<UUID, Map>) => {
+          dataSetIds.forEach((dataSetId :UUID) => {
+            const dataSet = getIn(state, [EDM, ORG_DATA_SETS, organizationId, dataSetId]) || Map();
+            mutableMap.set(dataSetId, dataSet);
```
anzioka, line 22: same comment about `getIn()` ifNotSetValue

src/components/typography/DataSetTitle.js
UnsungHero97: I'll remove the TODO.

src/core/permissions/reducers/getDataSetPermissionsPageReducer.js
UnsungHero97: It does have an initial state, set in `src/core/permissions/reducers/index.js`

```
[GET_DATA_SET_PERMISSIONS_PAGE]: RS_INITIAL_STATE,
```

src/core/permissions/sagas/getDataSetPermissionsPage.js
UnsungHero97: I don't think this will work. `getOrgDataSetsFromMeta` is the first thing that needs to happen in order for `selectOrgDataSets` to actually get things, after which `getPermissions` can happen.

If I were to combine in one `if` block, I would have to do `selectOrgDataSets` twice.

src/core/redux/selectors/selectMyKeys.js
UnsungHero97: It's possible that the key is set and the value is `undefined` so the "not set value" param wouldn't apply. This avoids that.

src/core/permissions/sagas/getDataSetPermissionsPage.js
UnsungHero97: Just really wanted to hammer home the fact that this is a boolean. I can change it, but I don't see the harm of keeping it.

src/core/edm/sagas/initializeOrganizationDataSet.js
UnsungHero97: It took me way too long to figure this out, so we should definitely make this easier. From ES:

> The query string “mini-language” is used by the Query string and by the q query string parameter in the search API.

I think what we actually need is a query string builder. For example, this is what I'm currently struggling with:

```
constraints: [
  { searchTerm: `NOT(entity.${propertyTypeIds.get(FQNS.OL_FLAGS)}:audit) AND ${query}` },
],
```

Maybe something like this would be worth considering to bring in:

https://github.com/sudo-suhas/elastic-builder

src/core/permissions/sagas/getDataSetPermissionsPage.js
anzioka: I don't understand. Are you saying that `selectOrgDataSets` depends on `getOrgDataSetsFromMeta`?

UnsungHero97: Yes.

src/core/edm/sagas/initializeOrganizationDataSet.js
UnsungHero97: Actually, https://github.com/sudo-suhas/elastic-builder won't work for us because it seems to generate what ES needs, which is different than what our api needs.

anzioka (approved):

solomontang (approved):

PR #847, opened 2021-02-05
APPS-2755: Switch to Chip component for Roles

bergieboy:
Screen Shot 2021-02-05 at 10 26 49 AM


src/containers/org/MemberRoleCard.js
```diff
@@ -55,32 +22,20 @@ const MemberRoleCard = ({
...
-      </Flex>
-      {
-        isUnassignable && (
-          <IconButton aria-label="unassign-role" onClick={handleClick}>
-            <FontAwesomeIcon fixedWidth icon={faTimes} />
-          </IconButton>
-        )
-      }
-    </RoleCard>
+    <Chip label={role.title} onClick={goToRolePath} onDelete={isUnassignable ? handleClick : undefined} />
```
solomontang, line 38: can you make this a proper link? Users should be able to middle click/open in a new tab. chance to drop dispatch as well.

```<Chip label={role.title} component="a" href={`#${rolePath}`} clickable onDelete={...} /> ```

**Review by solomontang — CHANGES_REQUESTED** [review]

**Review by solomontang — APPROVED** [review]

**Review by giuliacampana — APPROVED** [review]

File: src/containers/org/MemberRolesContainer.js [hunk 2]
```diff
@@ -23,6 +24,10 @@ import {
   paginationReducer,
 } from '../../utils/stateReducers/pagination';
 
+const Flex = styled.div`
+  display: flex;
+`;
```
UnsungHero97, line 29: `Flex` isn't really doing anything. You can achieve the same result without the extra component by using a `div` directly.

```
<div>
  {
    pagedRoles.map(...)
  {
</div>
```

Additionally, this results in chips being right next to each other... a little space wouldn't hurt:

Screen Shot 2021-02-08 at 3 11 24 PM

src/containers/org/MemberRolesContainer.js
```diff
@@ -97,16 +102,18 @@ const MemberRolesContainer = ({
...
-              onClick={handleOpen}
-              organizationId={organizationId}
-              role={role}
-              isUnassignable={isOwner} />
-        ))
-      }
+      <Flex>
+        {
+          pagedRoles.map((role) => (
+            <MemberRoleCard
```
UnsungHero97, line 108: No change required. It's not a card anymore, consider renaming.

src/containers/org/MemberRoleCard.js
```diff
@@ -66,21 +30,12 @@ const MemberRoleCard = ({
...
-          </IconButton>
-        )
-      }
-    </RoleCard>
+    <Chip
+        clickable
+        component="a"
+        href={`#${rolePath}`}
+        label={role.title}
+        onDelete={isUnassignable ? handleClick : undefined} />
```
UnsungHero97, line 38: No change required, just a thought. The logic in this component now looks very similar to the logic in `RoleChipsList` and `RoleOverflowPopover`. This seems like a good candidate for a reusable component, something like `RoleChip`, that can be used in `MemberRolesContainer` as well as in `PeopleTable`.

UnsungHero97 (changes requested):

src/containers/org/MemberRolesContainer.js
bergieboy: so I'm gonna keep the Flex and add a margin to the chips with it.

src/containers/org/MemberRoleChip.js
```diff
@@ -0,0 +1,44 @@
...
+import { Chip } from 'lattice-ui-kit';
+import type { Role, UUID } from 'lattice';
+
+import { ORG_ID_PARAM, ORG_ROLE, ROLE_ID_PARAM } from '../../core/router/Routes';
+
+type Props = {
+  onClick :(role :Role) => void;
+  organizationId :UUID;
+  role :Role;
+  authorized :boolean;
```
anzioka, line 14: please sort

anzioka (approved):

src/containers/org/people/components/RoleChipsList.js
```diff
@@ -55,8 +55,7 @@ const RoleChipsList = ({
   };
 
   const { id } = getUserProfile(member);
-  const getHandleDelete = (role :Role) => (e :SyntheticEvent<HTMLElement>) => {
-    e.preventDefault();
+  const getHandleDelete = (role :Role) => {
```
solomontang, line 58: update this function name to just handleDelete since it is no longer used to create a handler function.

solomontang (changes requested):

anzioka (approved):

solomontang (approved):

PR #863, opened 2021-02-10
APPS-2745 Better Error Handling for Org Settings Page

bergieboy:
Screen Shot 2021-02-16 at 12 54 19 PM


giuliacampana (dismissed):

src/containers/org/settings/OrgSettingsContainer.js
```diff
@@ -197,28 +201,44 @@ const OrgSettingsContainer = ({
...
+          isUnauthorized
+            ? (
+              <Typography color="error" variant="overline">
+                You are not authorized to view database credetials.
+              </Typography>
+            )
+            : (
+              <>
+                <div>
+                  <Typography component="h2" variant="body2">DATABASE NAME</Typography>
```
UnsungHero97, line 214: The request to get the database name only checks if the user has READ on the org, so I believe any member will be able to view the database name. However, to get the integration account, you must have OWNER on the org, so you'll have to split these up.

UnsungHero97 (changes requested):

src/containers/org/settings/OrgSettingsContainer.js
bergieboy: Well, the database name comes through in the IntegrationDetails, which is no populated if the `getIntegrationDetails` call fails. That's why I hid it.

UnsungHero97: The database name is a separate request being made in the same saga. The saga currently fails if either request fails, which needs to change.

bergieboy: ok, moved the return of integration details to the finally block

src/containers/org/settings/OrgSettingsContainer.js
```diff
@@ -14,12 +14,13 @@ import {
...
   Typography,
 } from 'lattice-ui-kit';
-import { useRequestState } from 'lattice-utils';
+import { useError, useRequestState } from 'lattice-utils';
 import { useDispatch, useSelector } from 'react-redux';
 import { RequestStates } from 'redux-reqseq';
 import type { Organization, UUID } from 'lattice';
 import type { RequestState } from 'redux-reqseq';
 
+import type { SagaError } from '../../../types';
```
UnsungHero97, line 23: Please run the import sorter.

src/containers/org/reducers/getOrganizationIntegrationDetailsReducer.js
```diff
@@ -31,10 +32,21 @@ export default function reducer(state :Map, action :SequenceAction) {
...
     },
-    FINALLY: () => state.deleteIn([GET_ORGANIZATION_INTEGRATION_DETAILS, action.id]),
+    FINALLY: () => {
+      const storedAction = state.getIn([GET_ORGANIZATION_INTEGRATION_DETAILS, action.id]);
+      if (storedAction) {
+        const organizationId :UUID = storedAction.value;
+        return state
+          .setIn([INTEGRATION_DETAILS, organizationId], action.value)
+          .deleteIn([GET_ORGANIZATION_INTEGRATION_DETAILS, action.id]);
+      }
```
UnsungHero97, line 48: Since you moved this to the `FINALLY` block, I don't think it needs to exist in the `SUCCESS` block.

UnsungHero97 (changes requested):

UnsungHero97 (approved):

UnsungHero97:
I accidentally merged this pull request prior to getting the second approval; I thought I was looking at a different pull request in a different project. Prior to merging, I had already reviewed this pull request and I believe it is safe. Additionally, this is merging into `develop` so we can confirm all is good on staging. As an admin, I don't think we need to revert the merge to get a second approval.

PR #999, opened 2021-04-09
APPS-2862: Type Org Name Before Deletion

bergieboy:
- Requires user to type the organizations' name before confirming deletion of Org. 
- If the deletion fails, the primary button switched to reset a request state function with text, 'Try Again'.
 
delete_success
delete_failure


solomontang (dismissed):

src/containers/org/components/OrgActionButton.js
```diff
@@ -94,6 +94,7 @@ const OrgActionButton = ({
 |}) => {
 
   const organizationId :UUID = (organization.id :any);
+  const organizationName :string = (organization.title :any);
```
UnsungHero97, line 97: Casting to `any` is not necessary.
```suggestion
  const organizationName :string = organization.title;
```

UnsungHero97 (dismissed):

UnsungHero97 (approved):

solomontang (approved):

PR #1027, opened 2021-04-27
search users by email

UnsungHero97:
This pull requests depends on https://github.com/openlattice/openlattice/pull/170

Search by email:

Screen Shot 2021-04-27 at 1 31 16 PM

Search with wildcard `*`:

Screen Shot 2021-04-27 at 1 31 28 PM


src/core/redux/selectors/selectUserSearchResults.js
```diff
@@ -0,0 +1,12 @@
...
+ */
+
+import { Map, getIn } from 'immutable';
+
+import { USERS, USER_SEARCH_RESULTS } from '../constants';
+
+export default function selectUserSearchResults() {
+
+  return (state :Map) :Map<string, Map> => getIn(state, [USERS, USER_SEARCH_RESULTS]) || Map();
+}
```
anzioka, line 12: annotation looks a little fuzzy

src/core/users/sagas/searchAllUsers.js
```diff
@@ -0,0 +1,73 @@
...
+
+  try {
+    yield put(searchAllUsers.request(action.id, action.value));
+
+    if (action.value === '*') {
+      throw new Error('search query cannot be "*"');
+    }
+    if (!isNonEmptyString(action.value)) {
+      throw new Error('search query cannot be empty');
+    }
```
anzioka, line 32: no change required, but I think it feels more natural for these two statements to be reversed (First verify that the value is a non-empty string, then verify the string is not of value *)

src/core/redux/selectors/selectUserSearchResults.js
UnsungHero97: What do you mean? The return value is a `Map` where the key is `string` and the value is `Map`.

src/core/users/sagas/searchAllUsers.js
UnsungHero97: Fixed.

src/core/redux/selectors/selectUserSearchResults.js
anzioka: ah ok! makes sense!

solomontang (approved):

bergieboy (approved):

giuliacampana (approved):

PR #1052, opened 2021-05-10
task/APPS-2838 check or uncheck all column permissions

giuliacampana:
Fixes:
- Don’t dispatch updatePermissions if the request body is going to be empty
- Show permissions lock icon (and disallow toggle click) when user doesn’t have owner on all columns in the dataset
- Ensure onlyNonPII spinner is not spinning when updating permissions for Data Set Object

locks:
Screen Shot 2021-05-10 at 9 40 50 AM

toggle not spinning:
c14fc2b7e8212bbb6a8dd24480a9bd6d


src/core/permissions/utils/computePermissionAssignments.js
```diff
@@ -26,17 +26,20 @@ function computePermissionAssignments(
...
   let isAssignedToOnlyNonPII = true;
+  const ownershipByColumn = [];
 
   dataSetColumns.forEach((column :Map<FQN, List>) => {
     const columnId :UUID = getPropertyValue(column, [FQNS.OL_ID, 0]);
     const propertyType :?PropertyType = maybePropertyTypes.get(columnId);
     const pii :boolean = propertyType?.pii || false;
     const key :List<UUID> = List([dataSetId, columnId]);
     const isOwner = myKeys.has(key);
+    ownershipByColumn.push(isOwner);
```
UnsungHero97, line 42: Using an array seems overkill since all you need to know is if there exists one case where `isOwner === true`. I think the following will work (and is slightly more efficient):

```
let isOwnerOfAtLeastOneColumn = false;
...
const isOwner = myKeys.has(key);
if (isOwner) {
  isOwnerOfAtLeastOneColumn = true;
  ...
}
```

src/core/permissions/utils/computePermissionAssignments.js
```diff
@@ -47,7 +50,9 @@ function computePermissionAssignments(
     }
   });
 
-  return { isAssignedToAll, isAssignedToOnlyNonPII };
+  const isOwnerOnAtLeastOneColumn = ownershipByColumn.reduce((result, isOwner) => isOwner, true);
```
UnsungHero97, line 53: This doesn't work. All this does is return the last value in the array. Separately, you don't need to do this if you switch to my suggestion above.

UnsungHero97 (changes requested):

UnsungHero97 (approved):

anzioka (approved):

PR #1084, opened 2021-06-02
APPS-2905 - Create Entity Data View

bergieboy:
Introduces UI to select data and explore its neighbors. 
- Searching an entity set for data remains the same
- Data can now be clicked on to open a modal that will display all fields in a grid view as well as all its neighbors. 
- Association and Neighbor Entity Sets are displayed as 'Chip' checkboxes. Selecting these entity sets will render neighbor tables below.
- There is a copy link button, which copies a link to a designated page for that entity. This is the same view as in the modal except the dataset information is displayed as breadcrumbs and the neighbor entities can now be clicked on to 'refocus' the page to that entity.
- Edge and Entity Data are merged into one table - edge data is marked with a light purple header and the word `edge` in parenthesis of the header.
- Added Dataset `title` to the `DataSetSearcchResultCard`
- Still need to add arrow functionality to click through search results in the modal.
ezgif com-gif-maker (7)
ezgif com-gif-maker (8)
Screen Shot 2021-06-02 at 6 28 12 PM


src/components/buttons/FlipButton.js
```diff
@@ -0,0 +1,58 @@
...
+
+type Props = {
+  children :Node;
+  className :string;
+  flip :boolean;
+  onClick ?:() => void;
+  size ?:string;
+};
+
+const FlipButton = (props :Props) => {
```
UnsungHero97, line 29: I think you might not need this component. There already exists a `Flip` component and this is how I use it:

```
<Flip flip={isOpenSearchOptions}>
  <IconButton aria-label="toggle search options" onClick={toggleSearchOptions}>
    <FontAwesomeIcon fixedWidth icon={faChevronDown} size="xs" />
  </IconButton>
</Flip>
```

You can see this in use in `OrgContainer`:

Screen Shot 2021-06-02 at 3 35 17 PM

Would this work for you?

giuliacampana (approved):

src/containers/explore/EntityDataContainer.js
```diff
@@ -0,0 +1,150 @@
...
+  organizationRoute,
+} :Props) => {
+  const dispatch = useDispatch();
+
+  const linkString = `https://openlattice.com/orgs/#${
+    Routes.ENTITY_DETAILS
+      .replace(Routes.DATA_SET_ID_PARAM, dataSetId)
+      .replace(Routes.ENTITY_KEY_ID_PARAM, entityKeyId)
+      .replace(Routes.ORG_ID_PARAM, organizationId)
+  }`;
```
UnsungHero97, line 83: Please don't hardcode `https://openlattice.com/orgs`.

src/containers/explore/EntityDataContainer.js
```diff
@@ -0,0 +1,150 @@
...
+  const entityData :?Map = useSelector((s) => s.getIn([EXPLORE, SELECTED_ENTITY_DATA], Map()));
+  const exploreEntityDataRS :?RequestState = useRequestState([EXPLORE, EXPLORE_ENTITY_DATA]);
+  const neighbors :?Map = useSelector((s) => s.getIn([EXPLORE, ENTITY_NEIGHBORS_MAP, entityKeyId], Map()));
+  const organization :?Organization = useSelector(selectOrganization(organizationId));
+
+  const name :string = getPropertyValue(dataSet, [FQNS.OL_DATA_SET_NAME, 0]);
+  const title :string = getPropertyValue(dataSet, [FQNS.OL_TITLE, 0]);
+
+  useEffect(() => {
+    if (isValidUUID(entityKeyId) && isValidUUID(dataSetId)) {
```
UnsungHero97, line 95: I'm thinking these might actually be more hurtful than helpful. If either is not a valid uuid, then nothing will happen, basically silent failure.  `exploreEntityData` calls `getEntityData`, which checks the id, so the saga will fail when it's invalid, and you can handle that error in the component. I think the same applies for `exploreEntityNeighbors`.

src/containers/explore/EntityDataContainer.js
```diff
@@ -0,0 +1,150 @@
...
+      dispatch(exploreEntityData({ entityKeyId, entitySetId: dataSetId }));
+      dispatch(exploreEntityNeighbors({ entityKeyId, entitySetId: dataSetId }));
+    }
+  }, [dispatch, entityKeyId, dataSetId]);
+
+  if (isPending(exploreEntityDataRS)) {
+    return (
+      <Spinner size="2x" />
```
UnsungHero97, line 103: Use `Spinner` from `'../../components'` which already does `size="2x"`.

src/containers/explore/EntityDataContainer.js
```diff
@@ -0,0 +1,150 @@
...
+
+  const linkString = `https://openlattice.com/orgs/#${
+    Routes.ENTITY_DETAILS
+      .replace(Routes.DATA_SET_ID_PARAM, dataSetId)
+      .replace(Routes.ENTITY_KEY_ID_PARAM, entityKeyId)
+      .replace(Routes.ORG_ID_PARAM, organizationId)
+  }`;
+
+  const dataSet :Map<FQN, List> = useSelector(selectOrgDataSet(organizationId, dataSetId));
+  const entityData :?Map = useSelector((s) => s.getIn([EXPLORE, SELECTED_ENTITY_DATA], Map()));
```
UnsungHero97, line 86: Please create a selector for this.

src/containers/explore/EntityDataContainer.js
```diff
@@ -0,0 +1,150 @@
...
+  const exploreEntityDataRS :?RequestState = useRequestState([EXPLORE, EXPLORE_ENTITY_DATA]);
+  const neighbors :?Map = useSelector((s) => s.getIn([EXPLORE, ENTITY_NEIGHBORS_MAP, entityKeyId], Map()));
```
UnsungHero97, line 88: Please create a selector for this.

src/containers/explore/EntityDataContainer.js
```diff
@@ -0,0 +1,150 @@
...
+`;
+
+type Props = {
+  dataSetDataRoute ?:string;
+  dataSetId :UUID;
+  entityKeyId :UUID;
+  isModal ?:boolean;
+  organizationId :UUID;
+  organizationRoute ?:string;
+};
```
UnsungHero97, line 66: No change required. I prefer to inline these because the tooltip actually shows the props. For example ...

Not inline:
Screen Shot 2021-06-07 at 7 30 20 PM

Inline:
Screen Shot 2021-06-07 at 7 30 10 PM

src/containers/explore/EntityDataContainer.js
```diff
@@ -0,0 +1,150 @@
...
+import { FQNS } from '../../core/edm/constants';
+import {
+  ENTITY_NEIGHBORS_MAP,
+  EXPLORE,
+  SELECTED_ENTITY_DATA
+} from '../../core/redux/constants';
+import { selectOrgDataSet, selectOrganization } from '../../core/redux/selectors';
+import { Routes } from '../../core/router';
+import { clipboardWriteText } from '../../utils';
+import { EntityDataGrid } from './components';
```
UnsungHero97, line 47: Please run the import sorter.

src/containers/explore/EntityDataContainer.js
```diff
@@ -0,0 +1,150 @@
...
+const { isValidUUID } = ValidationUtils;
+
+const FlexContainer = styled.div`
+  display: flex;
+  justify-content: space-between;
+  margin-bottom: 30px;
+`;
+
+type Props = {
+  dataSetDataRoute ?:string;
```
UnsungHero97, line 60:
```suggestion
  dataSetDataRoute :string;
```

src/containers/explore/EntityDataContainer.js
```diff
@@ -0,0 +1,150 @@
...
+  dataSetId :UUID;
+  entityKeyId :UUID;
+  isModal ?:boolean;
+  organizationId :UUID;
+  organizationRoute ?:string;
```
UnsungHero97, line 65:
```suggestion
  organizationRoute :string;
```

src/containers/explore/EntityDataContainer.js
```diff
@@ -0,0 +1,150 @@
...
+        </AppContentWrapper>
+      </>
+    );
+  }
+  return null;
+};
+
+EntityDataContainer.defaultProps = {
+  organizationRoute: '',
+  dataSetDataRoute: '',
```
UnsungHero97, line 146:
```suggestion
```

src/containers/explore/EntityDataContainer.js
UnsungHero97, line 103:
```suggestion
      <AppContentWrapper>
        <Spinner size="2x" />
      </AppContentWrapper>
```

Spinner is not centered:

Screen Shot 2021-06-07 at 7 36 34 PM

src/containers/explore/EntityDataContainer.js
```diff
@@ -0,0 +1,150 @@
...
+          }
+          <FlexContainer>
+            <Typography gutterBottom variant="h1">{entityKeyId}</Typography>
+            <LinkButton
+                color="default"
+                isDisabled={!linkString.length}
+                onClick={() => clipboardWriteText(linkString)}>
+              Get Link
+            </LinkButton>
+          </FlexContainer>
```
UnsungHero97, line 128: The button doesn't look great on smaller screens. 

Screen Shot 2021-06-07 at 7 43 28 PM

It behaves a little better if you use the existing `SpaceBetweenGrid` component.

```suggestion
          <SpaceBetweenGrid>
            <Typography gutterBottom variant="h1">{entityKeyId}</Typography>
            <LinkButton
                color="default"
                isDisabled={!linkString.length}
                onClick={() => clipboardWriteText(linkString)}>
              Get Link
            </LinkButton>
          </SpaceBetweenGrid>
```

src/containers/explore/EntityNeighborsContainer.js
```diff
@@ -0,0 +1,171 @@
...
+  margin: 15px 0;
+`;
+
+type Props = {
+  isModal :boolean;
+  neighbors :Map;
+  organizationId :UUID;
+};
+
+const EntityNeighborsContainer = ({ isModal, neighbors, organizationId } :Props) => {
```
UnsungHero97, line 37: Not sure exactly where to add this comment. I think I found an edge case. When `/neighbors/advanced/ids` returns nothing, then `EntityNeighborsContainer` renders nothing. It should instead show some kind of message to indicate the lack of neighbors.

Screen Shot 2021-06-07 at 7 49 53 PM

UnsungHero97 (changes requested):

src/containers/org/OrgRouter.js
```diff
@@ -303,6 +321,7 @@ const OrgRouter = () => {
 
     return (
       <Switch>
+        <Route exact path={Routes.ENTITY_DETAILS} render={renderDataSetDataDetailsContainer} />
```
UnsungHero97, line 324: I'm not quite sure where the bug is, but I can't view the non-modal version of `EntityDataContainer`. If I click the "Get Link" button to copy the url and then paste and navigate to it, then `EntityDataContainer` doesn't get rendered.

UnsungHero97 (changes requested):

src/containers/explore/EntityDataContainer.js
```diff
@@ -0,0 +1,150 @@
...
+                <CrumbLink to={dataSetDataRoute}>{title || name}</CrumbLink>
+                <CrumbItem>{entityKeyId}</CrumbItem>
+              </Crumbs>
+            )
+          }
+          <FlexContainer>
+            <Typography gutterBottom variant="h1">{entityKeyId}</Typography>
+            <LinkButton
+                color="default"
+                isDisabled={!linkString.length}
```
UnsungHero97, line 124: `isDisabled` will never be true because `linkString` will always have length.

UnsungHero97 (changes requested):

src/containers/explore/EntityDataContainer.js
bergieboy: What's the best way to handle this?

src/containers/explore/EntityDataContainer.js
bergieboy: the crumbs are only displayed on the stand-alone page, so these are actually optional arguments. No need to pass them in the modal.

src/containers/explore/EntityDataContainer.js
bergieboy: the crumbs are only displayed on the stand-alone page, so these are actually optional arguments. No need to pass them in the modal.

src/containers/explore/EntityDataContainer.js
bergieboy: see above

src/containers/explore/EntityDataContainer.js
UnsungHero97: Use `window.location`:

https://developer.mozilla.org/en-US/docs/Web/API/Window/location
https://developer.mozilla.org/en-US/docs/Web/API/Location

Screen Shot 2021-06-10 at 12 02 01 PM

src/core/redux/selectors/selectSelectedEntityData.js
```diff
@@ -0,0 +1,13 @@
+/*
+ * @flow
+ */
+
+import { Map } from 'immutable';
+
+import { EXPLORE, SELECTED_ENTITY_DATA } from '../constants';
+
+export default function selectEntityNeighborsMap() {
```
UnsungHero97, line 9:
```suggestion
export default function selectSelectedEntityData() {
```

src/core/redux/selectors/index.js
```diff
@@ -15,6 +15,7 @@ export { default as selectCurrentAuthorization } from './selectCurrentAuthorizat
 export { default as selectCurrentRoleAuthorizations } from './selectCurrentRoleAuthorizations';
 export { default as selectDataSetPermissionsPage } from './selectDataSetPermissionsPage';
 export { default as selectDataSetSchema } from './selectDataSetSchema';
+export { default as selectSelectedEntityData } from './selectSelectedEntityData';
```
UnsungHero97, line 18: Please sort.

src/core/redux/selectors/index.js
```diff
@@ -37,6 +38,7 @@ export { default as selectSearchHits } from './selectSearchHits';
 export { default as selectSearchPage } from './selectSearchPage';
 export { default as selectSearchQuery } from './selectSearchQuery';
 export { default as selectSearchTotalHits } from './selectSearchTotalHits';
+export { default as selectEntityNeighborsMap } from './selectEntityNeighborsMap';
```
UnsungHero97, line 41: Please sort.

src/containers/explore/components/EntityDataModal.js
```diff
@@ -0,0 +1,90 @@
...
+  isVisible :boolean;
+  onClose :() => void;
+  organizationId :UUID;
+|}) => {
+  const exploreEntityDataRS :?RequestState = useRequestState([EXPLORE, EXPLORE_ENTITY_DATA, entityKeyId]);
+  const exploreEntityNeighborsRS :?RequestState = useRequestState([EXPLORE, EXPLORE_ENTITY_NEIGHBORS, entityKeyId]);
+  const reducedRequestState :?RequestState = reduceRequestStates([exploreEntityDataRS, exploreEntityNeighborsRS]);
+
+  const header = (
+    <StyledModalHeader textTitle={`Examining ${dataSetName}: ${dataSetId}`} onClose={onClose} />
```
UnsungHero97, line 63: Close button doesn't work.

```suggestion
    <StyledModalHeader textTitle={`Examining ${dataSetName}: ${dataSetId}`} onClickClose={onClose} />
```

src/containers/explore/components/EntityDataModal.js
```diff
@@ -0,0 +1,90 @@
...
+const { isPending, reduceRequestStates } = ReduxUtils;
+
+const StyledModalHeader = styled(ModalHeader)`
+  background: ${NEUTRAL.N600};
+  text-align: center;
+
+  h1 {
+    color: white;
+  }
+`;
```
UnsungHero97, line 36: Figma has centered text and a white icon.

Screen Shot 2021-06-10 at 8 31 52 PM

src/containers/explore/components/EntityDataRow.js
```diff
@@ -0,0 +1,65 @@
...
+import { Routes } from '../../../core/router';
+import { goToRoute } from '../../../core/router/actions';
+
+const { NEUTRAL } = Colors;
+const { getEntityKeyId } = DataUtils;
+
+const StyledTableRow = styled.tr`
+  background-color: transparent;
+  border-bottom: 1px solid ${NEUTRAL.N100};
+  ${(props) => !props.isModal && StyleUtils.getHoverStyles};
```
UnsungHero97, line 22:
```suggestion
  ${(props) => !props.isModal && StyleUtils.getHoverStyles};
  
  &:last-child {
    border-bottom: none;
  }
```

... to get rid of that extra 1px border

Screen Shot 2021-06-10 at 8 45 50 PM

UnsungHero97 (changes requested):

src/containers/explore/EntityNeighborsTable.js
```diff
@@ -0,0 +1,177 @@
...
+
+  /*
+   * NOTE: performance issue against prod
+   * /explore/entityData/77f1b8d0-9c75-4e52-8176-eb6913a74669/f7cc0000-0000-0000-8000-0000000070ab
+   */
+
+  const dispatch = useDispatch();
+  const [tableData, setTableData] = useState([]);
+  const [tablePage, setTablePage] = useState(0);
+  const [neighborsIndex, setNeighborsIndex] = useState(0);
```
UnsungHero97, line 66:
```suggestion
  const [neighborsIndex, setNeighborsIndex] = useState(1);
```

UnsungHero97 (changes requested):

src/containers/explore/EntityNeighborsTable.js
```diff
@@ -0,0 +1,177 @@
...
+  neighbors,
+  organizationId
+} :Props) => {
+  const associationDataSetId :UUID = get(associationDataSet, 'id');
+  const dataSetId :UUID = get(dataSet, 'id');
+
+  /*
+   * NOTE: performance issue against prod
+   * /explore/entityData/77f1b8d0-9c75-4e52-8176-eb6913a74669/f7cc0000-0000-0000-8000-0000000070ab
+   */
```
solomontang, line 61: Are there any updates on this issue?

src/containers/explore/EntityNeighborsTable.js
```diff
@@ -0,0 +1,177 @@
...
+          let entityData = dataSetData.get(neighborEntityKeyId).set('id', `${neighborEntityKeyId}-${index}`);
+          associationData.get(associationEntityKeyId).forEach((value, key) => {
+            entityData = entityData.set(`${key}_edge`, value);
+          });
+          return entityData;
+        }).toJS(); // TODO: avoid .toJS()
+      setTableData(newTableData);
+    }
+  // NOTE: leaving out "dataSetData" from depedency array because it tends to cause infinite renders
+  // TODO: figure out how to avoid the infinite renders when "dataSetData" is passed
```
solomontang, line 122: This is because dataSetData is computed from selected values. you should instead be moving the computation inside the effect, rather than in the selector. I think this is an issue throughout. 

Whatever computation the selectors are doing should be broken out into external functions to be used inside these effects. We can also use reselect, which memoizes selectors.

src/core/edm/utils/useEntitySets.js
```diff
@@ -0,0 +1,57 @@
...
+const { isNonEmptyArray, isNonEmptyString } = LangUtils;
+const { EntitySet } = Models;
+
+type IdsOrNames =
+  | UUID[]
+  | string[]
+  | Set<UUID>
+  | Set<string>;
+
+// OPTIMIZE / MEMOIZE
```
solomontang, line 27: This would be a great candidate to try using https://github.com/reduxjs/reselect

src/core/edm/utils/useEntityTypePropertyTypes.js
```diff
@@ -0,0 +1,46 @@
...
+  EDM,
+  ENTITY_TYPES,
+  ENTITY_TYPES_INDEX_MAP,
+  PROPERTY_TYPES,
+  PROPERTY_TYPES_INDEX_MAP,
+} from '../../redux/constants';
+
+const { EntityType, FQN, PropertyType } = Models;
+const { isValidUUID } = ValidationUtils;
+
```
solomontang, line 21: This would be a great candidate to try using https://github.com/reduxjs/reselect

src/containers/explore/components/EntityDataGrid.js
```diff
@@ -0,0 +1,145 @@
...
+  });
+
+  return (
+    <>
+      <DataGrid>
+        { items.slice(0, 8) }
+        { showFull && items.slice(8) }
+      </DataGrid>
+      {
+        !!items.slice(8).length && (
```
UnsungHero97, line 128: A thought on this: `!!items.slice(8).length`. I'm sure this works but it's not very intuitive / immediately obvious, which I think is important when reading conditions like this. In addition, the sole purpose of this expression is to evaluate a boolean and I think this is a wasteful approach since `slice()` will create a copy of the array which just gets thrown away. Finally, you're already doing `items.slice(8)` above; why not reuse it?

I think the following is cleaner 

```
const hiddenItems = items.splice(8);

return (
  <>
    <DataGrid>
      {items}
      {showFull && hiddenItems}
    </DataGrid>
    {
      hiddenItems.length > 0 && (
```

src/containers/explore/components/EntityDataGrid.js
```diff
@@ -0,0 +1,145 @@
...
+          <InfoText>
+            <Typography variant="subtitle1">See</Typography>
+            <CountText variant="button">{` ${items.slice(8).length} `}</CountText>
```
UnsungHero97, line 131: You're doing another `slice()` here. This would become `hiddenItems.length`.

UnsungHero97 (changes requested):

src/containers/explore/EntityNeighborsTable.js
```diff
@@ -0,0 +1,177 @@
...
+  const dataSetId :UUID = get(dataSet, 'id');
+
+  /*
+   * NOTE: performance issue against prod
+   * /explore/entityData/77f1b8d0-9c75-4e52-8176-eb6913a74669/f7cc0000-0000-0000-8000-0000000070ab
+   */
+
+  const dispatch = useDispatch();
+  const [tableData, setTableData] = useState([]);
+  const [tablePage, setTablePage] = useState(0);
```
UnsungHero97, line 65:
```suggestion
  const [tablePage, setTablePage] = useState(1);
```

src/containers/explore/EntityNeighborsTable.js
```diff
@@ -0,0 +1,177 @@
...
+  const [neighborsIndex, setNeighborsIndex] = useState(1);
+  const fetchEntitySetDataRS :?RequestState = useRequestState([DATA, FETCH_ENTITY_SET_DATA, dataSetId]);
+
+  const neighborToAssociationEKIDs :Map = useMemo(() => (
+    Map().withMutations((mutableMap) => {
+      if (neighbors) {
+        neighbors
+          .slice(neighborsIndex, neighborsIndex + MAX_HITS_10)
+          .forEach((neighbor :Map) => mutableMap
+            .set(get(neighbor, 'neighborId'), get(neighbor, 'associationId')));
```
UnsungHero97, line 75: Just checking ... is this meant to be a one-to-one map?

src/containers/explore/EntityNeighborsContainer.js
```diff
@@ -0,0 +1,177 @@
...
+    }
+  };
+
+  const associationEntitySetChips = associationEntitySetsMap.valueSeq().map((associationEntitySet :Map) => {
+    const associationEntitySetName = associationEntitySet.get('title') || associationEntitySet.get('name');
+    const associationEntitySetId = associationEntitySet.get('id');
+    return (
+      <Checkbox
+          checked={!visibleOptions.get(associationEntitySetId, Set()).isEmpty()}
+          label={associationEntitySetName}
```
UnsungHero97, line 92: To fix this: `Warning: Each child in a list should have a unique "key" prop.`

```suggestion
          label={associationEntitySetName}
          key={associationEntitySetId}
```

src/containers/explore/EntityNeighborsContainer.js
```diff
@@ -0,0 +1,177 @@
...
+  });
+
+  const neighborDataSetChips = visibleOptions.valueSeq().flatten().toSet().map((neighborESID :UUID) => {
+    const neighborDataSet = dataSetsMap.get(neighborESID, Map());
+    const neighborDataSetName = neighborDataSet.get('title') || neighborDataSet.get('name');
+    const neighborDataSetId = neighborDataSet.get('id');
+    return (
+      <Checkbox
+          checked={visibleNeighbors.includes(neighborDataSetId)}
+          label={neighborDataSetName}
```
UnsungHero97, line 106: To fix this: `Warning: Each child in a list should have a unique "key" prop.`

```suggestion
          label={neighborDataSetName}
          key={neighborDataSetId}
```

src/containers/explore/EntityNeighborsContainer.js
```diff
@@ -0,0 +1,177 @@
...
+    );
+  });
+
+  const visibleNeighborTables = visibleOptions.entrySeq().map(([associationESID, neighborESIDs]) => {
+    const associationEntitySet = associationEntitySetsMap.get(associationESID, Map());
+    const associationEntitySetName = associationEntitySet.get('title') || associationEntitySet.get('name');
+    const associationEntitySetId = associationEntitySet.get('id');
+    return visibleNeighbors.some((neighborESID) => neighborESIDs.includes(neighborESID))
+      && (
+        <>
```
UnsungHero97, line 119: This needs to have `key` to avoid the `Each child in a list should have a unique "key" prop` warning.

```suggestion
        <Fragment key={associationEntitySetId}>
```

src/containers/explore/EntityNeighborsContainer.js
```diff
@@ -0,0 +1,177 @@
...
+          <br />
+          <Typography variant="h4">
+            { associationEntitySetName }
+          </Typography>
+          {
+            neighborESIDs.map((neighborESID) => {
+              const neighborDataSet = dataSetsMap.get(neighborESID, Map());
+              const neighborDataSetId = neighborDataSet.get('id');
+              return visibleNeighbors.includes(neighborDataSetId) && (
+                <>
```
UnsungHero97, line 129: This needs to have `key` to avoid the `Each child in a list should have a unique "key" prop` warning.

```suggestion
                <Fragment key={neighborDataSetId}>
```

src/containers/explore/EntityNeighborsContainer.js
```diff
@@ -0,0 +1,177 @@
...
+  const visibleNeighborTables = visibleOptions.entrySeq().map(([associationESID, neighborESIDs]) => {
+    const associationEntitySet = associationEntitySetsMap.get(associationESID, Map());
+    const associationEntitySetName = associationEntitySet.get('title') || associationEntitySet.get('name');
+    const associationEntitySetId = associationEntitySet.get('id');
+    return visibleNeighbors.some((neighborESID) => neighborESIDs.includes(neighborESID))
+      && (
+        <>
+          <br />
+          <Typography variant="h4">
+            { associationEntitySetName }
```
UnsungHero97, line 122: I think this should also use `DataSetTitle` because it is an entity set. Or maybe neither should use it.

Screen Shot 2021-06-23 at 4 38 41 PM

UnsungHero97 (changes requested):

src/core/edm/utils/useEntitySets.js
UnsungHero97: https://github.com/openlattice/lattice-orgs/pull/1084#discussion_r654156521

src/core/edm/utils/useEntityTypePropertyTypes.js
UnsungHero97: https://github.com/openlattice/lattice-orgs/pull/1084#discussion_r654158157

src/core/redux/selectors/selectOrgDataSets.js
```diff
@@ -18,8 +25,13 @@ export default function selectOrgDataSets(organizationId :UUID, dataSetIds ?:UUI
...
           dataSetIds.forEach((dataSetId :UUID) => {
-            const dataSet = getIn(state, [EDM, ORG_DATA_SETS, organizationId, dataSetId]) || Map();
-            mutableMap.set(dataSetId, dataSet);
+            const entitySetIndex :number = state.getIn([EDM, ENTITY_SETS_INDEX_MAP, dataSetId], -1);
+            if (entitySetIndex >= 0) {
+              const dataSet :?EntitySet = state.getIn([EDM, ENTITY_SETS, entitySetIndex]);
+              if (dataSet && dataSet.id) {
+                mutableMap.set(dataSetId, dataSet.toImmutable());
+              }
+            }
```
UnsungHero97, line 34: Please revert this change. The app has switched to the "data set" model and I believe this is no longer correct. In fact, I think we need to completely remove getting/storing entity sets (probably property types too), i.e. `[EDM, ENTITY_SETS]` and `[EDM, ENTITY_SETS_INDEX_MAP]`.

UnsungHero97 (changes requested):

src/containers/explore/EntityNeighborsContainer.js
```diff
@@ -0,0 +1,177 @@
...
+  min-height: 500px;
+`;
+
+const SectionWrapper = styled.section`
+  margin: 15px 0;
+`;
+
+type Props = {
+  isModal :boolean;
+  neighbors :Map;
```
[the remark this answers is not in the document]
bergieboy: I feel like it shouldn't be optional, now that I'm looking at it. I'll modify it to be the other way around.

src/containers/explore/EntityNeighborsContainer.js
bergieboy: Talked to shireen, and we're gonna do this for now. She's gonna think if there is a better solution.
Screen Shot 2021-06-25 at 12 53 35 PM

src/containers/explore/EntityNeighborsTable.js
bergieboy: this comment was pulled in from another file. removed.

src/containers/explore/EntityNeighborsTable.js
bergieboy: yeah

src/core/redux/selectors/selectOrgDataSets.js
bergieboy: so, to be clear, the getEntitySet and getEntitySets reducers should be refactored to map to ORG_DATA_SETS rather than ENTITY_SETS?

bergieboy: so just to be clear, `getEntitySets` and `getEntitySet` should be mapped to `ORG_DATA_SETS` now?

src/containers/explore/EntityNeighborsTable.js
bergieboy: looking into this now.

bergieboy: what computation are you referring to that needs to be made in the effect?

solomontang: `selectOrgEntitySetData` creates a new Map every time because it transforms the values kept in the store into some other format. This transformation needs to either happen here in the effects, or the selector should be memoized.

@UnsungHero97 has attempted to work around this by placing object hashes as dependencies. This not only requires additional calculations, but also incurs a different type of warning where the listed dependency is not used.

Either relocating the transform into the effect or memoizing the selector will resolve both of these issues.

src/core/redux/selectors/selectOrgDataSets.js
UnsungHero97: No. `getEntitySet` and `getEntitySets` are ok as they are, it's just that this `selectOrgDataSets` needs to be reverted to what it was before.

src/containers/explore/EntityNeighborsTable.js
```diff
@@ -0,0 +1,177 @@
...
+    })
+  ), [neighbors, neighborsIndex]);
+
+  const totalNeighbors :number = useMemo(() => (
+    neighbors ? neighbors.count() : 0
+  ), [neighbors]);
+
+  // OPTIMIZE: no need to compute this on every render
+  const associationPropertyTypes :PropertyType[] = useEntityTypePropertyTypes(associationDataSet.get('entityTypeId'));
+  const neighborPropertyTypes :PropertyType[] = useEntityTypePropertyTypes(dataSet.get('entityTypeId'));
```
UnsungHero97, line 86: You'll have to switch to the selector.

UnsungHero97 (changes requested):

UnsungHero97 (dismissed):

UnsungHero97 (approved):

anzioka (approved):

PR #1175, opened 2021-07-26
Add ID to Dataset Detail View

bergieboy:
Screen Shot 2021-07-26 at 1 35 42 PM


src/containers/org/OrgDataSetContainer.js
```diff
@@ -126,7 +129,14 @@ const OrgDataSetContainer = ({
...
                 <div>
-                  <Typography variant="h1">{title || name}</Typography>
+                  <Typography gutterBottom variant="h1">{title || name}</Typography>
+                  <Typography variant="subtitle1">DATASET ID</Typography>
+                  <ActionsGrid align={{ v: 'center' }} fit>
+                    <Pre>{dataSetId}</Pre>
+                    <CopyButton
+                        aria-label="copy organization id"
+                        onClick={() => clipboardWriteText(dataSetId)} />
+                  </ActionsGrid>
```
UnsungHero97, line 139: Can you move this below `SpaceBetweenGrid`? The reason is that the menu button is getting centered with the height of the title + this new stuff and it doesn't look as good, in my opinion.  Additionally, it will look consistent across pages (data set, org, role, member).

UnsungHero97 (changes requested):

UnsungHero97 (approved):

solomontang (approved):

PR #1176, opened 2021-07-26
Allow for Bulk Permissions on Data Sets

bergieboy:


src/containers/permissions/assign-permissions-to-data-set/AssignPermissionsToDataSetModalBody.js
```diff
@@ -71,8 +72,9 @@ const AssignPermissionsToDataSetModalBody = ({
     .join(', ');
 
   const confirmText = assignPermissionsToAllProperties
-    ? `Please confirm you want to assign ${permissions} to "${targetDataSetTitle}" and all its columns.`
-    : `Please confirm you want to assign ${permissions} to "${targetDataSetTitle}".`;
+    ? `Please confirm you want to assign ${permissions} to "${targetDataSetTitles.join(', ')}" `
+      + 'datasets and all of their columns.'
```
UnsungHero97, line 76: This could become pretty unreadable as data set names can be fairly large ...

> Please confirm you want to assign OWNER, READ, WRITE to "Shiproom: DemoAddresses", "Shiproom: DemoSentencedWith", "Shiproom: DemoArrestedBy", "Shiproom: DemoChargeEvents" datasets and all of their columns

I'm thinking we should turn this into a `<ul>`:

```
Please confirm you want to assign OWNER, READ, WRITE to the following data sets and all of their columns:
  - Shiproom: DemoAddresses
  - Shiproom: DemoSentencedWith
  - Shiproom: DemoArrestedBy
```

src/containers/permissions/assign-permissions-to-data-set/AssignPermissionsToDataSetModalBody.js
```diff
@@ -71,8 +72,9 @@ const AssignPermissionsToDataSetModalBody = ({
...
+    : `Please confirm you want to assign ${permissions} to "${targetDataSetTitles.join(', ')}" datasets.`;
```
UnsungHero97, line 77: Another thought here ... I think it might be useful to specifically call out that the columns will not get permissions.

src/utils/getDataSetsKeys.js
```diff
@@ -0,0 +1,23 @@
...
+import { List, Map, get } from 'immutable';
+import type { UUID } from 'lattice';
+
+export default function getDataSetsKeys(
+  dataSets :Map<UUID, Map>,
+  dataSetsColumns :Map<UUID, Map<UUID, Map>>,
+) :List<List<UUID>> {
+
+  return List().withMutations((keys :List<List<UUID>>) => {
+    dataSets.forEach((dataSet, dataSetId) => {
```
UnsungHero97, line 14: Just being nitpicky here ... you only care about the key, so `.keySeq().forEach()` should be a little better because the `dataSet` object doesn't get brought into scope.

src/utils/getDataSetsKeys.js
```diff
@@ -0,0 +1,23 @@
...
+      const dataSetColumns = dataSetsColumns.get(dataSetId);
+      keys.push(List([dataSetId]));
+      dataSetColumns.forEach((column :Map) => {
```
UnsungHero97, line 17: Same comment. I believe `dataSetColumns` is a map from column id to column, so I think you can use `keySeq()` here as well.

UnsungHero97 (changes requested):

solomontang (approved):

UnsungHero97 (approved):

giuliacampana (approved):

PR #1198, opened 2021-08-20
add tags to data set metadata

solomontang:
- Allow users to select from a list of common metadata tags for datasets or type to create their own.
- prototype for https://jira.openlattice.com/browse/APPS-3125
image
image


giuliacampana (approved):

src/containers/org/OrgDataSetContainer.js
```diff
@@ -74,6 +75,12 @@ const CountBadge = styled(Badge)`
   margin-right: 5px;
 `;
 
+const StyledChip = styled(Chip)`
+  color: ${NEUTRAL.N700};
+  background: ${NEUTRAL.N100};
+  font-weight: 600;
+`;
```
bergieboy, line 82: should we just switch the default chip to match these colors?

bergieboy (approved):

src/containers/org/OrgDataSetContainer.js
solomontang: hmm dunno the designs are in consistent in that regard.

anzioka (approved):

src/containers/org/OrgDataSetContainer.js
UnsungHero97: I agree, we should use the chip colors. I've told Chris to standardize around that.

src/containers/org/OrgDataSetContainer.js
```diff
@@ -116,6 +124,9 @@ const OrgDataSetContainer = ({
   const collabCount = collaborationsByDataSetId.size;
   const collabTabText = isSuccess(getCollabWithDataSetRS) ? `Collaborations (${collabCount})` : 'Collaborations';
 
+  const tags = metadata.get(TAGS, List());
+  const tagLabel = `Protections: ${tags.join(', ')}`;
```
UnsungHero97, line 128: We'll eventually want to support any freeform text for tags, in addition to standardized ones, so I don't think we should have the `Protections: ` prefix.

UnsungHero97 (changes requested):

src/utils/constants/index.js
```diff
@@ -52,3 +52,89 @@ export const EDIT_TITLE_DESCRIPTION_UI_SCHEMA = {
...
+    label: 'Welfare and Institutions Code',
+    options: [
+      { label: 'WIC', value: 'WIC' },
+      { label: 'WIC § 4515', value: 'WIC § 4515' },
+      { label: 'WIC § 10850(f)', value: 'WIC § 10850(f)' },
+    ]
+  },
+];
+
+export const EDIT_TITLE_DESCRIPTION_FLAGS_DATA_SCHEMA = {
```
UnsungHero97, line 88:
```suggestion
export const EDIT_TITLE_DESCRIPTION_TAGS_DATA_SCHEMA = {
```

src/utils/constants/index.js
```diff
@@ -52,3 +52,89 @@ export const EDIT_TITLE_DESCRIPTION_UI_SCHEMA = {
...
+          title: 'Description',
+        },
+      },
+      required: ['title', 'description'],
+    },
+  },
+
+};
+
+export const EDIT_TITLE_DESCRIPTION_FLAGS_UI_SCHEMA = {
```
UnsungHero97, line 121:
```suggestion
export const EDIT_TITLE_DESCRIPTION_TAGS_UI_SCHEMA = {
```

UnsungHero97 (approved):

anzioka (approved):

PR #1210, opened 2021-08-25
Bugfix/modal reset state stuff

bergieboy:


UnsungHero97 (dismissed):

src/containers/collaborations/sagas/addDataSetsToCollaboration.js
```diff
@@ -55,6 +64,14 @@ function* addDataSetsToCollaborationWorker(action :SequenceAction) :Saga<*> {
     );
     if (responseError) throw responseError;
 
+    if (dataSetIdsToLoad.length) {
```
UnsungHero97, line 67: No change required. I feel like this will always be true ... you are adding every single id from `dataSetIdsToLoad`.

PR #1244, opened 2021-09-10
remove username/password and change default name

solomontang:


src/containers/org/utils/generateIntegrationConfig.js
```diff
@@ -22,8 +22,8 @@ datalakes:
   url: "jdbc:postgresql://atlas.openlattice.com:30001/${TEMPLATE_CONSTANTS.ORG_ID}?ssl=true&sslmode=require"
   driver: "org.postgresql.Driver"
   latticeLogger: true
-  username: "${TEMPLATE_CONSTANTS.ORG_USERNAME}"
-  password: "${TEMPLATE_CONSTANTS.ORG_PASSWORD}"
+  username: "<INSERT_USERNAME_HERE>"
+  password: "<INSERT_PASSWORD_HERE>"
```
UnsungHero97, line 26: No change required. Why not make this change inside the `TEMPLATE_CONSTANTS` object? Also, are these creds different from the creds above? If they are, we should specify `_ORG_` here.

solomontang: @kengie what's the rationale for not wanting the user/pass included in the generated template?

kengie: yeah, they could be different from the creds above. One for the source and one for the destination database.

kengie (approved):

UnsungHero97 (approved):

PR #1273, opened 2021-09-23
Allow for partial string search on datasets

bergieboy:


src/core/search/sagas/searchOrganizationDataSets.js
```diff
@@ -60,11 +60,19 @@ function* searchOrganizationDataSetsWorker(action :SequenceAction) :Saga<WorkerR
...
-    const constraints = [{
-      constraints: [{
-        searchTerm: query,
-      }],
-    }];
+    /*
+      allows for partial strings so titles that have no spaces and have not been
+      tokenized in elasticsearch as separate words are also returned in results
+    */
+    const searchQuery = (query !== '*') ? `*${query}*` : query;
```
UnsungHero97, line 67: How about ...

```suggestion
    const searchQuery = !query.includes('*') ? `*${query}*` : query;
```

UnsungHero97 (changes requested):

src/core/search/sagas/searchOrganizationDataSets.js
UnsungHero97: Actually, let's flip that around to avoid the `!`:

```
const searchQuery = query.includes('*') ? query : `*${query}*`;
```

UnsungHero97 (approved):

solomontang (approved):

PR #1286, opened 2021-09-30
Feature/apps 3119 bulk revoke roles from members

solomontang:
- Add bulk action menu option to revoke roles from multiple selected members. Flow is same as bulk assignment of roles

image
image

image

### DIFF
--- a/src/containers/org/components/RevokeRolesFromMembersBody.js
+++ b/src/containers/org/components/RevokeRolesFromMembersBody.js
@@ -1,0 +1,93 @@
+/*
+ * @flow
+ */
+
+import React, { useState } from 'react';
+
+import { Map } from 'immutable';
+import { useRequestState, useStepState } from 'lattice-utils';
+import { useDispatch } from 'react-redux';
+import type { Role, UUID } from 'lattice';
+import type { RequestState } from 'redux-reqseq';
+
+import BulkRoleSelectionStep from './BulkRoleSelectionStep';
+import ConfirmStepBody from './ConfirmStepBody';
+
+import { ORGANIZATIONS } from '../../../core/redux/constants';
+import { getUserProfile } from '../../../utils';
+import { REVOKE_ROLES_FROM_MEMBERS, revokeRolesFromMembers } from '../actions';
+
+const RESET_ACTIONS = [REVOKE_ROLES_FROM_MEMBERS];
+
+type Props = {
+  members :Map;
+  onClose :() => void;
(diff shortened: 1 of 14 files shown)


src/containers/org/components/RevokeRolesFromMembersBody.js
```diff
@@ -0,0 +1,93 @@
...
+  const dispatch = useDispatch();
+  const [step, stepBack, stepNext] = useStepState(2);
+  const requestState :?RequestState = useRequestState([ORGANIZATIONS, REVOKE_ROLES_FROM_MEMBERS]);
+  const [selectedRoles, setSelectedRoles] = useState(Map());
+
+  const handleSelectRole = (role :Role) => {
+    const { id } = role;
+    const newSelection = selectedRoles.has(id)
+      ? selectedRoles.remove(id)
+      : selectedRoles.set(id, role);
```
UnsungHero97, line 44: I recommend using `Set<UUID>` instead of `Map<UUID, Role>` because you're not actually using the role object for anything.

src/containers/org/sagas/revokeRolesFromMembers.js
```diff
@@ -0,0 +1,84 @@
...
+      members :Map<string, Map>;
+    |} = action.value;
+
+    if (!isValidUUID(organizationId)) {
+      throw new Error('organizationId must be a valid UUID');
+    }
+
+    const allRevokeRequests = [];
+    members.forEach((member :Map, memberId :string) => {
+      roles.forEach((role :Role, roleId :UUID) => {
```
UnsungHero97, line 48: Related to my other comment ... you don't need the role itself, just the id.

src/containers/org/sagas/revokeRolesFromMembers.js
```diff
@@ -0,0 +1,84 @@
...
+        const assignment = call(removeRoleFromMemberWorker, removeRoleFromMember({
+          memberId,
+          organizationId,
+          roleId,
+        }));
+        allRevokeRequests.push(assignment);
```
UnsungHero97, line 54:
```suggestion
        const revokeCall = call(removeRoleFromMemberWorker, removeRoleFromMember({
          memberId,
          organizationId,
          roleId,
        }));
        allRevokeRequests.push(revokeCall);
```

UnsungHero97 (approved):

anzioka (approved):

PR #1289, opened 2021-10-01
src/common/constants, src/common/types, src/common/utils

UnsungHero97:


src/common/utils/getSecurablePrincipalId.js
```diff
@@ -0,0 +1,38 @@
...
+  let principalId;
+
+  if (isImmutable(user)) {
+    securablePrincipalClass = getIn(user, [PRINCIPAL, AT_CLASS]);
+    principalId = getIn(user, [PRINCIPAL, ID]);
+  }
+  else {
+    securablePrincipalClass = _get(user, [PRINCIPAL, AT_CLASS]);
+    principalId = _get(user, [PRINCIPAL, ID]);
+  }
```
anzioka, line 31: `if..else` block here seems unnecessary since`getIn` works with both immutables and plain JS objects.

src/common/utils/isEntitySet.js
```diff
@@ -0,0 +1,15 @@
...
+import _get from 'lodash/get';
+import { get, isImmutable } from 'immutable';
+
+import { DATA_SET_TYPE, DataSetTypes } from '../constants';
+
+export default function isEntitySet(dataSet :any) :boolean {
+
+  return isImmutable(dataSet)
+    ? get(dataSet, DATA_SET_TYPE) === DataSetTypes.ENTITY_SET
+    : _get(dataSet, DATA_SET_TYPE) === DataSetTypes.ENTITY_SET;
```
anzioka, line 14: same comment here about immutable's `get`

src/common/utils/getSecurablePrincipalId.js
UnsungHero97: It does not work for custom objects, like the `Principal` class.

src/common/utils/isEntitySet.js
UnsungHero97: Same response.

solomontang (commented):
I'm curious to hear why you're seeking this change. Managing imports seems like work your IDE should be able to help with.

Are you also considering keeping all sagas in one file? All actions? Reducers? React components?

src/common/utils/getSecurablePrincipalId.js
anzioka: `user` here is not an instant of `Principal` class though, right?

UnsungHero97:
> I'm curious to hear why you're seeking this change. Managing imports seems like work your IDE should be able to help with.

Atom does not automatically import and I haven't been able to find a plugin to do that. That's not what I'm struggling with though. The reason I’m doing this is because I keep finding myself being confused about where to import constants/utils from, and I thought it would be easier if we pull it all together into a single place. Furthermore, there have been numerous times where duplicate constants are declared; some of these cases have been caught during review, others have not.

> Are you also considering keeping all sagas in one file?

No. I don't want to put everything in one file. I don't want one giant file with a bunch of logic. We used to do that with sagas and reducers, and I'm actively moving away from that pattern. What I am trying to do is switch to importing from one place, which is how we currently import common components, because I think importing from one place makes things easier and cleaner. In other words, `index.js` is meant to export everything. 

Yes, the constants are in one file because I don't see an immediate benefit to splitting them up into many different files, but that's easy to do and still aligns with my goal with this pull request.

> All actions?

Maybe. I haven't set aside time to think through this one.

> Reducers?

No.

> React components?

No. However, maybe it makes sense to move the `components` folder inside `common`. Or alternatively, `common` doesn't need to exist and everything moves up one level. I like `common` though.

anzioka (approved):

solomontang (approved):

